PR #215, opened 2015-01-14
Fix the success msgs in the personal settings page

nickvergessen:
Currently the success message is not displayed:
owncloud_-_2015-01-14_17 24 14

Fixed:
owncloud_-_2015-01-14_17 26 23

@LukasReschke

### DIFF
--- a/controller/settings.php
+++ b/controller/settings.php
@@ -135,7 +135,7 @@ class Settings extends Controller {
 		return new DataResponse(array(
 			'status'	=>'success',
 			'data'		=> array(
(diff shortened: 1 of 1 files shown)


LukasReschke:
:+1:

DeepDiver1975:
famous issue :+1:

DeepDiver1975:
I really wonder if we shouldn't better adopt t() to return a true string :question:

nickvergessen:
I also wondered if there is any usecase where someone needs the OC_L10N_String object.
I guess we could add another method for that case.

PR #216, opened 2015-01-15
log if a notification mail could not be set

MorrisJobke:
Maybe helps spotting problems.

as discussed with @nickvergessen in IRC

### DIFF
--- a/lib/backgroundjob/emailnotification.php
+++ b/lib/backgroundjob/emailnotification.php
@@ -38,6 +38,9 @@ class EmailNotification extends \OC\BackgroundJob\TimedJob {
 	/** @var \OCP\IConfig */
 	protected $config;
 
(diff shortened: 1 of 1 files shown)


karlitschek:
good idea :+1:

lib/backgroundjob/emailnotification.php
```diff
@@ -92,6 +96,7 @@ protected function runStep($limit) {
 			if (!isset($userEmails[$user])) {
 				// The user did not setup an email address
 				// So we will not send an email :(
+				$this->logger->debug("Couldn't send notification mail to user '" . $user . "'", ['app' => 'activity']);
```
nickvergessen: Add that the email adress needs to be set up?

MorrisJobke: Sounds reasonable

PR #222, opened 2015-02-03
replace deprecated config calls with proper method calls

MorrisJobke:
Ref owncloud/core#12260

### DIFF
--- a/rss.php
+++ b/rss.php
@@ -31,8 +31,7 @@ if (!\OCP\User::isLoggedIn()) {
 		exit;
 	}
 
-	$preferences = new \OC\Preferences(\OC_DB::getConnection());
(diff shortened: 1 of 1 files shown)


MorrisJobke:
See https://github.com/owncloud/core/blob/master/lib/private/preferences.php#L132

Raydiation:
What about also killing legacy OC_Preferences? Only usage is in lib/private/installer.php but im only using core + calendar + news + notes + contacts

MorrisJobke:
@Raydiation https://github.com/owncloud/core/pull/13869 ;)

nickvergessen:
will merge, as soon as stable8 branches are created

PR #223, opened 2015-02-05
Directly use JobList service from container

nickvergessen:


nickvergessen:
@LukasReschke +1 ? ;)

LukasReschke:
Makes sense :+1:

PR #234, opened 2015-02-13
Remove the status attribute and display a success message instead

### LINKED ISSUE #212: DataResponses from controller/settings.php should not use status attribute
But instead it should rely on the HTTP status and headers

nickvergessen:
- [x] Requires https://github.com/owncloud/core/pull/14208

Fixes #212

### DIFF
--- a/controller/settings.php
+++ b/controller/settings.php
@@ -133,7 +133,6 @@ class Settings extends Controller {
 		);
 
 		return new DataResponse(array(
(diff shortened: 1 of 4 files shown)


nickvergessen:
@MorrisJobke I assume you used this to test owncloud/core#14208
If so, mind +1-ing so we can merge this?

MorrisJobke:
:+1:

PR #238, opened 2015-02-17
Use json_encode/json_decode instead of (un)serialize() for the parameters

### LINKED ISSUE #237: Use `json_decode` instead of `unserialize`
`unserialize` can lead to all kind of problems when we suddenly happen to have user input passed to it…

nickvergessen:
Fix #237 

@LukasReschke


LukasReschke:
:heart:

LukasReschke:
Tested, works. Data is still there and converted :+1:

PR #239, opened 2015-02-19
Unify the names of the stream template files

nickvergessen:
Just some renaming, to make it easier to realise which file is what part.
The `list.php` is now an empty body instead of a list of activities, `activity.box.php` referes to the boxes from the oC6 design, now it's just one sentence mostly, etc.
So the file names refered to old behaviour.

@LukasReschke


LukasReschke:
Mhm… No unit tests for the controllers? :see_no_evil:

LukasReschke:
Still :+1: - can you create an issue to create unit tests for those?

PR #242, opened 2015-02-19
Add unit tests to cover all constructors and classes at least once

nickvergessen:
Some basic tests to include all files into the test suite (apart from rss.php)

@LukasReschke


LukasReschke:
:+1:

PR #244, opened 2015-02-23
Use a valid timezone name when no timezone is set

nickvergessen:
We create a TimeZone object from the value directly:
https://github.com/owncloud/activity/blob/83865de1b1285f4c4540f4047fba86cfbee4ee6d/lib/mailqueuehandler.php#L196
See log in https://github.com/owncloud/core/issues/14176

@LukasReschke

### DIFF
--- a/lib/backgroundjob/emailnotification.php
+++ b/lib/backgroundjob/emailnotification.php
@@ -101,7 +101,7 @@ class EmailNotification extends \OC\BackgroundJob\TimedJob {
 			}
 
 			$language = (isset($userLanguages[$user])) ? $userLanguages[$user] : $default_lang;
-			$timezone = (isset($userTimezones[$user])) ? $userTimezones[$user] : '';
+			$timezone = (isset($userTimezones[$user])) ? $userTimezones[$user] : 'UTC';
 			$this->mqHandler->sendEmailToUser($user, $userEmails[$user], $language, $timezone, $data);
 		}
 


nickvergessen:
Should backport to 8.0 @karlitschek

LukasReschke:
Mhm. We set the timezone in https://github.com/owncloud/core/blob/e87ada86d16043402f64ae45d75995a6cffb7c51/lib/base.php#L551 with `date_default_timezone_set`, might be better to use `date_default_timezone_get` then here?

Anyways, both approaches are valid and a backport shouldn't hurt :+1:

deMattin:
LukasReschke wrote:
> Mhm. We set the timezone in https://github.com/owncloud/core/blob/e87ada86d16043402f64ae45d75995a6cffb7c51/lib/base.php#L551 with date_default_timezone_set, might be better to use date_default_timezone_get then here?
>
> Anyways, both approaches are valid and a backport shouldn't hurt :+1:

I think a hard coded UTC really hurts!
The use of server time is much better because mostly the server uses the same timezone as the user using this server.
And why do we have to set the values in php.ini if we do not use this?

Another possible and better (best?) way would be to set a default timezone in config.php or use the log-timezone as default timezone.

But how to set user-timezone (calender timezone doesn't work for this)?
The lack of setting the user timezone nor in config-panel of user nor in admin-panel is another thing that REALLY hurts because it seems to be used in oc-code and can't reliable set and checked.
And for now the only way is to do so in database directly.

deMattin:
https://github.com/owncloud/activity/pull/245

PR #245, opened 2015-02-24
[Stable8] Use a valid timezone name when no timezone is set

nickvergessen:
Backport of #244 
@LukasReschke

### DIFF
--- a/lib/backgroundjob/emailnotification.php
+++ b/lib/backgroundjob/emailnotification.php
@@ -101,7 +101,7 @@ class EmailNotification extends \OC\BackgroundJob\TimedJob {
 			}
 
 			$language = (isset($userLanguages[$user])) ? $userLanguages[$user] : $default_lang;
-			$timezone = (isset($userTimezones[$user])) ? $userTimezones[$user] : '';
+			$timezone = (isset($userTimezones[$user])) ? $userTimezones[$user] : 'UTC';
 			$this->mqHandler->sendEmailToUser($user, $userEmails[$user], $language, $timezone, $data);
 		}
 


karlitschek:
:+1:

deMattin:
If you use UTC here as fallback, Activity-Mails are send in my OC 8.0.0.8 with UTC.
1. Where is the user-timezone "$userTimezones[$user]" set?
The only way to set a timezone in User-Config is in calender-app. But this is not a basic app and so it has to be in another place. But to set it here doen't help at all.
2. Would't it be better to use the server-timezone set in config.php if there is no user-timezone found?
I testet:
$timezone = (isset($userTimezones[$user])) ? $userTimezones[$user] : $logtimezone;
instead of: $timezone = (isset($userTimezones[$user])) ? $userTimezones[$user] : '';
but this doesn't seem to work.
$logtimezone already has a fallback to UTC if not set in config.php, I think.

nickvergessen:
The timezone is set when a user logs himself into the oc instance, so that should happen to most users at some point.

nickvergessen:
And btw, a comment on one issue would have been enough.
There is no need to post three almost equal comments

deMattin:
Ok, Sorry for the cross commenting.
The cross postings are not really equal - they are more supplementary.and should have been merged in one place by me. I became lost in git. ;)

Let's go on here ...

1. It seems to be a login "cookie problem", that the user timezone hasn't been set. After Upgrade the cookies are still valid and there is no need to re-login and so there has no timezone variable been created for this users.
I tested this with my main account and a logout/login sets the user-timezone-variable in database.
Thanks for this information.

2. What do you think about my suggestion to use the server logtimezone as fallback instead of using the hardcoded UTC, as I proposed in https://github.com/owncloud/activity/pull/244?
In my opinion this will be the best and most simple solution to set a better fitting timezone if the user has no timezone set in database.

nickvergessen:
1. hmm, sounds like something we should consider. Mind opening a ticket in core?
2. See #248

deMattin:
In #248 you set it to default timezone. For me, this is also a good solution, if servers php.ini is fully configured.

All solutions have advantages and disadvantages.
Let me write down some thoughts:

Use of logtime variable in config.php:
++ (only) used in owncloud
++ parameter documented in owncloud (but comment has to be enhanced in config-default.php, if also used as user default timezone)
++ has fallback to UTC if not set by admin
++ works with php.ini default timezone not being set
-- not possible to set another logtime timzone as the server timezone (only relevant for server/user in other countries/timezone as the admin uses)

Use of (php) default timezone:
++ already mentioned in admin documentation as a value to be set (https://doc.owncloud.org/server/8.0/admin_manual/installation/source_installation.html#configuration-notes-to-php-ini-files)
-- value has to be set in the correct php.ini config-file
-- fails, if no default timezone is set!
-- who reads manual installation notes if using a repository? ;)

Use a new config variable in config.php:
++ Transparent and easy to understand way
-- A bit overdone for this (yet) not often used variable

Using php.ini and as fallback logtime variable:
-- not very transparent and confusing dependency

Am I missing something important?

__So for me using logtime as user default timezone seems to be the best solution.__

Enhancing php-code of owncloud and debugging is really a pain for me because I don't know php very well.
For testing this problem, I changed something and had to see, if cronjob does his job the right way.
I'm not to lazy to enhance code the right way, but am not really capable ...

So could you please do this, if you agree with my thoughts?

PR #246, opened 2015-02-25
Testcoverage for email stuff

nickvergessen:
Fix #251 
Master port of #248

PR #247, opened 2015-02-25
[Stable8] Do not try to use empty values

nickvergessen:
@LukasReschke @DeepDiver1975 

Backporting the fix from #246 to stable8


LukasReschke:
:+1:

PR #252, opened 2015-03-09
Use displayname as user recipient in emails

nickvergessen:
Backport #251

Tests are in master only #246

@DeepDiver1975 @karlitschek approve backport?


karlitschek:
Great. Please backport after 8.0.2

PR #255, opened 2015-03-13
Issue/254 increase parameter length

### LINKED ISSUE #254: 255 characters is too short for subject and message parameters
Most activities contain a uid + filepath as subject parameters.
The uid is 64 characters as per `oc_users`.`uid`
File path is `oc_filecache`.`name` (250) + `oc_filecache`.`path` (4000), so 4250 characters in worst case.

However the planned move/rename activity might have 2 file paths in max length, so I think we should change the column from VAR(255) to TEXT, to avoid problems with apps.

The existing broken entries should be deleted on update:
```sql
SELECT *
FROM `oc_activity`
WHERE `subjectparams` NOT LIKE '%}' OR `messageparams` NOT LIKE '%}'
```

Error log from broken entries:
```
unserialize(): Error at offset 248 of 256 bytes at /var/www/apps/activity/lib/grouphelper.php#69

nickvergessen:
Fix #254 

@LukasReschke should fix your problem with the Log spam about unserialize() as well as speeding up the update from 8.0 to 8.1 (15+ minutes down to 10 seconds here)


karlitschek:
We have to be careful with schema changes. Changing type prevents users from doing the db schema change before the owncloud code is upgraded.
I prefere if you would introduce a new column for that.
:-1:

nickvergessen:
> We have to be careful with schema changes. Changing type prevents users from doing the db schema change before the owncloud code is upgraded.

??? Mind elaborating? we change from string to text here, so that shouldn't be very incompatible.

> I prefere if you would introduce a new column for that.

If I use a new column we loose all the existing data, so this is not really the best idea, or we keep the broken column as well, which should be quite confusing.

karlitschek:
@nickvergessen Got it. This should most likely work. Sorry for the noise. In general we have to try to add new columns instead of changing existing ones to enable zero downtime upgrading. 
This specific case should work of course.

DeepDiver1975:
> This specific case should work of course.

Is there a test case which handles migration? clob and Oracle rings some :bell: to me .....

nickvergessen:
@karlitschek removed the column type change, so this should be good to go then.
The change itself was already tested by @LukasReschke iirc?

LukasReschke:
Fine for me :+1:

karlitschek:
:+1:

PR #261, opened 2015-03-16
Use a closure for the navigation entry, to speed up webdav and ajax call...

nickvergessen:
- [x] Depends on https://github.com/owncloud/core/pull/14931


karlitschek:
makes sense :+1:

LukasReschke:
:+1:

PR #269, opened 2015-03-24
Limit the number of selected email entries per user

### LINKED ISSUE #267: Activity email can potentially get infinite long
Should we trim the text after ~200 entries or something like that?

nickvergessen:
Fix #267 

@karlitschek as discussed. The limit has been set to 200.


karlitschek:
as discussed :+1:

PR #271, opened 2015-03-25
Limit file activities to favorites

nickvergessen:
- [x] Depends on #276 
- [x] https://github.com/owncloud/core/pull/15196

PR #276, opened 2015-03-30
Fix compilation

nickvergessen:
- [x] Use the correct app name for sharing events (required for #271 + https://github.com/owncloud/core/pull/15312 )
- [x] Do not wrap the activity with a link, when the activity contains one
- [x] Allow extensions to filter the 'all' filter (required for #271)
- [x] Hide the navigation entries when they are just duplicates or empty

@LukasReschke


karlitschek:
makes sense :+1:

LukasReschke:
:+1:

PR #285, opened 2015-04-24
Fix link to RSS

LukasReschke:
The defined route does not exists and throws an error 500

@nickvergessen


nickvergessen:
What's the use of tests, if they expect the wrong thing :(

:+1:

PR #294, opened 2015-05-11
Issue 293 add active class li element

### LINKED ISSUE #293: Changing navigation.php to match files-app css class .active
Hey there,

I was theming my owncloud the last days and there's one thing where I stuck. Pictures are worth than a thousand words, so:

How it should be
files

How it is
activities

You see on the second pic, that the nav elements (Alle Aktivitäten) color isn't matching the appcontent container.

My CSS should be okay, but the problem seems to be that in your app the `.active` class is set within the `<a>` element instead of the `<li>` element (you also could blame the files-app but this app is build in^^).
As I was figuring out why this happens, I changed the source code to bind that `<li>` to the `.active` class. This was working (more or less), but to handle this via theming isn't that cool. So I ask you, if you could change it to match the class handling of the navigation of the files-app. :-)

A big THX in advance :-)

nickvergessen:
Fix #293

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -11,7 +11,7 @@ $(function(){
 				return;
 			}
 
-			this.navigation.find('a[data-navigation=' + this.filter + ']').removeClass('active');
+			this.navigation.find('a[data-navigation=' + this.filter + ']').parent().removeClass('active');
 			this.currentPage = 0;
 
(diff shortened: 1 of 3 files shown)


karlitschek:
makes sense :+1:

PR #295, opened 2015-05-11
Fix the detection of self-actions on files activities

### LINKED ISSUE #289: Exclude your own actions in the RSS activity feed
It does not seem possible to filter **out** (exclude) your own activities in the RSS activity feed. I would assume that by unticking "List your own actions in the stream" one would get only other's activities, yet this seems to have no effect (I still see "You created...", "You deleted...", "You changed....."). I've just upgraded to 8.0.3 but still no difference.

What I am trying to achieve is getting an RSS feed (or email notification) of only other users / anonymous uploads / downloads etc, and entirely ignore my own actions. 

screenshot-2015-05-05 21-28-49

nickvergessen:
Fix #289

### DIFF
--- a/lib/fileshooks.php
+++ b/lib/fileshooks.php
@@ -290,7 +290,7 @@ class FilesHooks {
 			return;
 		}
 
-		$selfAction = substr($subject, -5) !== '_self';
+		$selfAction = $user === $this->currentUser;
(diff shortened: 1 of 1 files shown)


karlitschek:
:+1:

nickvergessen:
@karlitschek we should backport this to activity app for 8.0.4

karlitschek:
please backport :+1:

PR #300, opened 2015-05-13
Increase the versions for the 8.1 release

nickvergessen:
The oC min. requirement is needed for:
* The navigation manager `\Closure` thing for the performance improvement
* The split between activity app logic and activities (which now live in the respective app)

So 8.1 activity app does not work with the old core. Therefor I also updated the major version for the activity app.

Sounds good? @DeepDiver1975


MorrisJobke:
:+1:

karlitschek:
:+1:

PR #306, opened 2015-06-17
Do not escape the translations, so the french can use single quotes j…

nickvergessen:
…ust fine

@LukasReschke 

@karlitschek should backport to stable8


LukasReschke:
As discussed :+1:

PR #308, opened 2015-06-23
Add Oracle-specific workaround for changing a column type

VicDeo:
Ref https://github.com/owncloud/core/issues/16812

"6.0.4 to 7.0.5 upgrade with Oracle as backend database"

### DIFF
--- a/appinfo/preupdate.php
+++ b/appinfo/preupdate.php
@@ -1,0 +1,28 @@
+<?php
+
+$installedVersion = \OC::$server->getConfig()
+				->getAppValue('activity', 'installed_version');
+
+if (version_compare($installedVersion, '1.1.1', '>=') && version_compare($installedVersion, '1.1.2', '<=')) {
+	$connection = \OC::$server->getDatabaseConnection();
+	$platform = $connection->getDatabasePlatform();
+	if ($platform->getName() === 'oracle') {
+		try {
+			$connection->beginTransaction();
+			$sql1 = 'ALTER TABLE `*PREFIX*activity` ADD `type_text` VARCHAR2(255) DEFAULT NULL';
+			$connection->executeQuery($sql1);
+			$sql2 = 'UPDATE `*PREFIX*activity` SET `type_text` = to_char(`type`)';
+			$connection->executeQuery($sql2);
+			$sql3 = 'ALTER TABLE `*PREFIX*activity` DROP COLUMN `type` cascade constraints';
+			$connection->executeQuery($sql3);
+			$sql4 = 'ALTER TABLE `*PREFIX*activity` RENAME COLUMN `type_text` TO `type`';
+			$connection->executeQuery($sql4);
+			$connection->commit();
+		} catch (\Doctrine\DBAL\DBALException $e) {
+			\OC::$server->getLogger()->warning(
+				"Oracle upgrade fixup failed: " . $e->getMessage(), 
+				['app' => 'activity']
+			);
+		}
+	}
+}


appinfo/preupdate.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+$installedVersion=OCP\Config::getAppValue('activity', 'installed_version');
+if (version_compare($installedVersion, '1.1.2', '=')) {
+	$connection = OC_DB::getConnection();
```
nickvergessen: `\OC::$server->getDatabaseConnection()`

appinfo/preupdate.php
```diff
@@ -0,0 +1,23 @@
...
+	$platform = $connection->getDatabasePlatform();
+	if ($platform->getName() === 'oracle') {
+		try {
+			$connection->beginTransaction();
+			$sql1 = 'ALTER TABLE `*PREFIX*activity` ADD `type_text` VARCHAR2(255) DEFAULT NULL';
+			\OC_DB::executeAudited($sql1, array());
```
nickvergessen: No `OC_DB`

appinfo/preupdate.php
```diff
@@ -0,0 +1,23 @@
...
+			$sql2 = 'UPDATE `*PREFIX*activity` SET `type_text` = to_char(`type`)';
+			\OC_DB::executeAudited($sql2, array());
```
nickvergessen: No `OC_DB`

appinfo/preupdate.php
```diff
@@ -0,0 +1,23 @@
...
+			$sql3 = 'ALTER TABLE `*PREFIX*activity` DROP COLUMN `type` cascade constraints';
+			\OC_DB::executeAudited($sql3, array());
```
nickvergessen: No `OC_DB`

appinfo/preupdate.php
```diff
@@ -0,0 +1,23 @@
...
+			$sql4 = 'ALTER TABLE `*PREFIX*activity` RENAME COLUMN `type_text` TO `type`';
+			\OC_DB::executeAudited($sql4, array());
```
nickvergessen: No `OC_DB`

nickvergessen:
Am I right, that this would need backporting all the way to stable7?

appinfo/preupdate.php
VicDeo: @nickvergessen it should reach stable7 ;)

VicDeo:
@nickvergessen I think so
cc @DeepDiver1975

appinfo/preupdate.php
```diff
@@ -0,0 +1,23 @@
...
+			$connection->commit();
+		} catch (\DatabaseException $e) {
+			\OCP\Util::writeLog('activity', "Oracle upgrade fixup failed: " . $e->getMessage(), \OCP\Util::WARN);
```
nickvergessen: `\OC::$server->getLogger()->warn()`

appinfo/preupdate.php
nickvergessen: Well then change it in the patch that is for stable7, but on master we are not going start using private and deprecated APIs

DeepDiver1975:
> Am I right, that this would need backporting all the way to stable7?

we have to as we need to fix migration on oracle for oc6 -> oc7

nickvergessen:
Okay, I will try to take this over from here on then

nickvergessen:
> Okay, I will try to take this over from here on then

Anyway, feel free to continue @VicDeo

DeepDiver1975:
> Anyway, feel free to continue @VicDeo

yes please @VicDeo - your task :speak_no_evil:

appinfo/preupdate.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+$installedVersion=OCP\Config::getAppValue('activity', 'installed_version');
+if (version_compare($installedVersion, '1.1.2', '=')) {
```
nickvergessen: should be a comparison "if lower than", not equals... Otherwise 1.1.1 will not run this update, but it should

appinfo/preupdate.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+$installedVersion=OCP\Config::getAppValue('activity', 'installed_version');
```
nickvergessen: `\OC::$server->getConfig()`

VicDeo: @nickvergessen OCP\Config is a _public_ API, isn't it? ;)

appinfo/preupdate.php
VicDeo: `\OCP\Util::writeLog` is not deprecated in master

appinfo/preupdate.php
nickvergessen: but deprecated

nickvergessen: https://github.com/owncloud/core/blob/master/lib/public/config.php#L46

nickvergessen:
Actually we don't need this file in stable8 and master?
Updating from 6 to 8 is not allowed?

VicDeo:
@nickvergessen updated in accordance with your comments

VicDeo:
@nickvergessen 
> Updating from 6 to 8 is not allowed?

Right.
But disabling Activity, updating 6 -> 7 -> 8 -> xxx and enabling Activity is allowed.

DeepDiver1975:
> But disabling Activity, updating 6 -> 7 -> 8 -> xxx and enabling Activity is allowed.

ouch - so true ... do we want to support such a scenario?

@karlitschek

nickvergessen:
Well okay, the let's say the activity app does not support a major version jump:
1.1.x > 1.2.x > 2.0.x
So that should work as well?

DeepDiver1975:
> Well okay, the let's say the activity app does not support a major version jump:
> 1.1.x > 1.2.x > 2.0.x
> So that should work as well?

well - the main problem here is that app upgrade are not triggered if they are not active - which is okay.
But this can result in the behavior as explained by @VicDeo in https://github.com/owncloud/activity/pull/308#issuecomment-114976039

Until we find a better mechanism to handle migrations more seemless (pssst owncloud/core#14851 :speak_no_evil: ) we might simply add the migration to every release branch - even if they are never executed?

nickvergessen:
So we also have to bring back the update.php which was dropped because of the normally unsupported update-way?
https://github.com/owncloud/activity/commit/24ab014e10223c9a6c4abe7b334ad4386f2e8573

I mean if someone encounters the problem, we can still say, hey please upload the files of 1.2.x (oC 8.0) first then update to that version and then update to 2.0 (oC 8.1)?

DeepDiver1975:
> So we also have to bring back the update.php which was dropped because of the normally unsupported update-way?
> 24ab014
> 
> I mean if someone encounters the problem, we can still say, hey please upload the files of 1.2.x (oC 8.0) first then update to that version and then update to 2.0 (oC 8.1)?

valid point as well ....

VicDeo:
I never argue for a long time (and don't use Oracle anyway).
Let those unsupported migrators die if it looks correct from your standpoint.

nickvergessen:
Yeah, lets just fix this in 6>7 then

PR #312, opened 2015-06-24
[stable7] Add Oracle-specific workaround for changing a column type

VicDeo:
Ref https://github.com/owncloud/core/issues/16812

### DIFF
--- a/appinfo/preupdate.php
+++ b/appinfo/preupdate.php
@@ -1,0 +1,23 @@
+<?php
+
+$installedVersion=OCP\Config::getAppValue('activity', 'installed_version');
+if (version_compare($installedVersion, '1.1.1', '>=') && version_compare($installedVersion, '1.1.2', '<=')) {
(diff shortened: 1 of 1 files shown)


appinfo/preupdate.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+$installedVersion=OCP\Config::getAppValue('activity', 'installed_version');
+if (version_compare($installedVersion, '1.1.1', '>=') && version_compare($installedVersion, '1.1.2', '<=')) {
```
nickvergessen: This looks wrong, all versions need this update?!

VicDeo: @nickvergessen nope, only those that have int `type` field (1.1.1 and 1.1.2)
Adding a `type` is no problem. The problem observed while changing this field from int to varchar.

nickvergessen: Ah, I see the type field was only added in 1.1.1

PR #321, opened 2015-07-02
Remove OCS id because the app is shipped

### LINKED ISSUE #320: Activity app in app store?
I need a little help on the decision here. Currently the activity app has a ocsid entry in the info.xml:
https://github.com/owncloud/activity/blob/master/appinfo/info.xml#L22
all the way down to stable7
https://github.com/owncloud/activity/blob/stable7/appinfo/info.xml#L16

However, in the app store there was only a dummy, so if it would have been used to update the app, it would have killed itself? Anyhow, @LukasReschke removed the dummy yesterday to avoid problems.

Now the question is, should we properly maintain the activity app in the appstore, so people can also update it when they are too lazy to update core with the next maintenance release? Or should we just remove the ocsid everywhere and say that it is only available via the normal packages.

@DeepDiver1975 @karlitschek

nickvergessen:
Fix #320


MorrisJobke:
:+1:

LukasReschke:
:+1:

PR #326, opened 2015-07-02
Find the federated cloud id in contacts for better readability

nickvergessen:
Following the sample of the share dropdown from owncloud/core#17165

Look for contacts with the cloud ID:

1. First line no contact entry, just trimming the remote into tipsy like in the file list
2. Contact is found, use name from there and tipsy is federated cloud id

Stream:
bildschirmfoto vom 2015-07-02 13 08 53

Feed: 
activity_feed_-_2015-07-02_13 08 42

@jancborchardt @schiesbn @karlitschek 

**NOTE:** For testing https://github.com/owncloud/core/pull/17335 (or only the commit https://github.com/owncloud/core/commit/ddb8cc23034ebd4ca5bc16a35ae9d18ac5fb8dac ) is required, because otherwise there is no activity that uses the parameter type "federated_cloud_id"


nickvergessen:
PS: old behaviour is to always display "admin@localhost/owncloud/master/core" like in the stream screenshot.

karlitschek:
:+1:

lib/parameterhelper.php
```diff
@@ -179,20 +192,84 @@ protected function prepareUserParam($param, $highlightParams) {
...
 
 		if ($highlightParams) {
 			$avatarPlaceholder = '';
 			if ($this->config->getSystemValue('enable_avatars', true)) {
 				$avatarPlaceholder = '<div class="avatar" data-user="' . $param . '"></div>';
 			}
-			return $avatarPlaceholder . '<strong>' . $displayName . '</strong>';
+			return $avatarPlaceholder . '<strong>' . Util::sanitizeHTML($displayName) . '</strong>';
 		} else {
 			return $displayName;
```
LukasReschke: Looks fishy. Previously $displayName was also sanitized. Care to elaborate on this change?

LukasReschke: (as well as for the others below)

nickvergessen: This is for `$highlightParams === false` so when we don't return HTML.
The problem is previously the RSS feed displayed `Foo&amp;bar shared a.txt with you`

nickvergessen:
Rebased due to https://github.com/owncloud/activity/pull/344

lib/parameterhelper.php
nickvergessen:
```irc
<nickvergessen> btw https://github.com/owncloud/activity/pull/326/files#r33854454 is okay with you?
<nickvergessen> I would  then rebase and merge it
<LukasReschke> 👍

PR #328, opened 2015-07-03
Return activities as JSON instead of sending HTML

### LINKED ISSUE #272: When multiple files/folders are grouped into one message, show multiple icons/thumbnails
Currently every action just has one icon or thumbnail. That way it always looks as if you just modified one thing.

Instead, when there were for example 7 images uploaded, it should show the thumbnails of those images. For the text above we have a limit of 5, after that it shows `and x more …`, so we can also cut off after the first 5 thumbnails and show a dot-dot-dot next to that.

cc @nickvergessen

nickvergessen:
Fix #272 

Required for the rightside bar #358 

For testing just go to the stream, switch between master and the branch and refresh the page. It should only display more previews, nothing else should change.

@PVince81 the ground work for #358


PVince81:
Minor issue: "today" is now writte lower case ? Is that intentional ?

PVince81:
Some of the spacings/paddings changed.
You can see it by opening the same activities on two tabs, one with master and one with this branch. Then switch the tabs quickly :wink: 

It's all minor, we can merge this first.

PR #331, opened 2015-07-04
Use the SVG icon like in the file list

### LINKED ISSUE #329: File type icons are blurry cause they do not use SVG
You can easily see this when zoomed, or when using a smartphone:
capture du 2015-07-04 00-17-06

@nickvergessen @georgehrke

nickvergessen:
Fix #329 

Using the same "hack" as the file list:
https://github.com/owncloud/core/blob/master/apps/files/lib/helper.php#L80

@jancborchardt 

we can then backport this until stable8 I think

### DIFF
--- a/lib/display.php
+++ b/lib/display.php
@@ -101,7 +101,9 @@ class Display {
 					))
 				);
 			} else {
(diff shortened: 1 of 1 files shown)


jancborchardt:
Thanks btw, sorry for missing the review here! :)

PR #333, opened 2015-07-16
Do not group remote users (they might be different users)

nickvergessen:
Non-local users are currently not distinguishable, so grouping them might
remove the information how many different users performed the same action.
So we do not group them anymore, until we found another solution.

Fixes owncloud/core#17569 

Although I think the problem is a different one, this helps to work around it. The grouping is useful and should be brought back, but first we need to be able to distinguish the different users.

@MorrisJobke please review

@karlitschek we shall backport this to 8.1 for the time being


MorrisJobke:
Tested and works with this patch :+1:

MorrisJobke:
@gig13 Mind testing this?

gig13:
@MorrisJobke 
I will test with OC8.0.5.

MorrisJobke:
> I will test with OC8.0.5.

@nickvergessen Also mind a backport to stable8?

karlitschek:
Great. :+1:  please backport

nickvergessen:
Should be easily backportable once confirmed

gig13:
@MorrisJobke @nickvergessen 
Let me know where to get the backport files, I tested with the ones above with OC8.0.5 and the Activity Stream just hung with the spinner.

nickvergessen:
stable8.1 #334 
stable8.0 #335

MorrisJobke:
@gig13 Backport for stable8 https://github.com/owncloud/activity/pull/335.patch

gig13:
@MorrisJobke @nickvergessen 
Working in stable8.0

PR #336, opened 2015-07-18
Revert back to the old method instead of the public one that tries to…

nickvergessen:
… read it

@enoch85 @deMattin please test

### DIFF
--- a/lib/display.php
+++ b/lib/display.php
@@ -91,7 +91,7 @@ class Display {
 			$tmpl->assign('previewLink', $this->getPreviewLink($activity['file'], $is_dir));
 
 			// show a preview image if the file still exists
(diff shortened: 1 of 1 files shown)


enoch85:
I can't find display.php in the lib directory...

nickvergessen:
That is in the lib/ of activity ;)

enoch85:
@nickvergessen *doh* 'Of course. :) Will test this later tonight.

enoch85:
@nickvergessen :+1:  on this! Tested and fixes the bug.

nickvergessen:
> Analysing /home/travis/build/owncloud/core/apps/activity/lib/display.php
 1 errors
    line   94: OC_Helper - static method call on private class
App is not compliant

This breaks the code checker, but until a public method that is suitable has been added/backported, nothing we can do.

PR #338, opened 2015-07-18
Add PHP 7 to test matrix

LukasReschke:


.travis.yml
```diff
@@ -4,6 +4,7 @@ php:
   - 5.5
   - 5.6
   - hhvm
+  - 7
```
nickvergessen: swap with hhvm [/stylenazi]

PR #344, opened 2015-07-22
Do not use deprecated Util::linkTo()

nickvergessen:
@LukasReschke please review


LukasReschke:
@owncloudPlus1Bot Do your duty.

PR #358, opened 2015-08-13
Activity tab in files sidebar

PVince81:
@nickvergessen FYI

Not sure what attributes the endpoint is going to return eventually.
Everything is a dummy at the moment.

- [x] REQUIRES https://github.com/owncloud/core/pull/18178
- [x] REQUIRES https://github.com/owncloud/core/pull/18290

### DIFF
--- a/js/activitycollection.js
+++ b/js/activitycollection.js
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2015
+ *
+ * This file is licensed under the Affero General Public License version 3
+ * or later.
+ *
+ * See the COPYING-README file.
+ *
+ */
+
+(function() {
+	/**
+	 * @class OCA.Activity.ActivityCollection
+	 * @classdesc
+	 *
+	 * Displays activity information for a given file
+	 *
+	 */
+	var ActivityCollection = OC.Backbone.Collection.extend(
+		/** @lends OCA.Activity.ActivityCollection.prototype */ {
+
+		/**
+		 * Id of the file for which to filter activities by
+		 *
+		 * @var int
+		 */
+		_objectId: null,
+
+		/**
+		 * Type of the object to filter by
+		 *
+		 * @var string
+		 */
+		_objectType: null,
+
+		model: OCA.Activity.ActivityModel,
+
+		initialize: function(options) {
+			if (options && options.fileId) {
+				this._fileId = options.fileId;
+			}
+		},
+
+		/**
+		 * Sets the object id to filter by or null for all.
+		 * 
+		 * @param {int} objectId file id or null
+		 */
+		setObjectId: function(objectId) {
+			this._objectId = objectId;
+		},
+
+		/**
+		 * Sets the object type to filter by or null for all.
+		 * 
+		 * @param {int} objectType file id or null
+		 */
+		setObjectType: function(objectType) {
+			this._objectType = objectType;
+		},
+
+		url: function() {
+			var query = {
+				page: 1,
+				filter: 'all'
+			};
+			var url = OC.generateUrl('apps/activity/activities/fetch');
+			if (this._objectId) {
+				query.objectid = this._objectId;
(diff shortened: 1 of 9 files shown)


PVince81:
For https://github.com/owncloud/core/issues/18164

karlitschek:
You guys rock!!  :-)

PVince81:
also requires https://github.com/owncloud/core/pull/18290

PVince81:
@nickvergessen I fixed the spinner issue.
Also I made ActivityCollection more generic to accept "objecttype" and "objectid".
Please implement the backend to make the filter work :smile: 

Currently it looks like this using the existing CSS from the page:
activity-nofilter

jancborchardt:
Nice! :) Of course the activities shown in there should only be the ones pertaining to that specific file. :)
In the screenshot above: Not the image addition or the folder creation – we need to keep it clean.

appinfo/app.php
```diff
@@ -49,3 +49,12 @@
 // Cron job for sending emails and pruning the activity list
 $c->getServer()->getJobList()->add('OCA\Activity\BackgroundJob\EmailNotification');
 $c->getServer()->getJobList()->add('OCA\Activity\BackgroundJob\ExpireActivities');
+
+// Needed for the files sidebar entry
+\OCP\Util::connectHook(
```
nickvergessen: Move to `FilesHooksStatic::register()`

PVince81:
@jancborchardt see previous comment

>Please implement the backend to make the filter work

PVince81:
- [ ] TODO: implement backend for the filter
- [ ] TODO: fix hook

js/activitycollection.js
```diff
@@ -0,0 +1,82 @@
...
+		 * @param {int} objectType file id or null
+		 */
+		setObjectType: function(objectType) {
+			this._objectType = objectType;
+		},
+
+		url: function() {
+			var query = {
+				page: 1,
+				filter: 'all'
```
nickvergessen: filter: 'filter'

PVince81: but only when objectid or objecttype are set ? or always ?

nickvergessen: all gives all, filter uses object id and type, see the following lines:
https://github.com/owncloud/activity/pull/363/files#diff-7842161f69cff3997d7b10126d8c368fR223

nickvergessen:
activity filter option is merged, event listener for loading js/css is merged

if you rebase we can test it all together

PVince81:
Will do

PVince81:
Rebased, fixed the hook and adapted.
It works now.

- [ ] TODO: JS unit tests

PVince81:
Oops, I missed some files

PVince81:
Should be good now.

PVince81:
- [ ] REQUIRES: https://github.com/owncloud/core/pull/18459 for the unit test loading

PVince81:
Unit tests added. But they will only run once https://github.com/owncloud/core/pull/18459 is merged in core.

PVince81:
See how difficult it is to load core's JS files in the right order: https://github.com/owncloud/activity/pull/358/files#diff-6e207d934280c969b0add642e133ee90R29

Fortunately we added "core/js/core.json" that lists the core JS files for that which can be reused here.

js/activitycollection.js
nickvergessen: this still needs updating?

PVince81: scroll down

PVince81: we only set the filter if either objectId or objectType are set. Else we fetch all (could be used later in other views)

PVince81:
This is ready for review @jancborchardt @nickvergessen

nickvergessen:
Works :+1:

jancborchardt:
Yup, nice! :)

Issue #360, opened 2015-08-14
Setup JS unit tests


PVince81:
Need to bring in a framework (probably karma or mocha) and have it run on Travis.

nickvergessen:
yes please, feel free to help :)

PVince81:
We're going to need it as the JS code is florishing even more :smile: 
I'll have a look later.

PVince81:
PR here https://github.com/owncloud/activity/issues/361

PR #361, opened 2015-08-14
Added JS unit test runner

PVince81:
Fixes https://github.com/owncloud/activity/issues/360

@nickvergessen 

Note that this is only the ground work. Unit tests to be written separately (at least for newly introduced code!)

Let's see if this runs properly on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ script:
 
   # Run Javascript unit tests
   - cd tests/js
-  - sh -c "if [ '$JSTESTS' != '1' ]; then npm install --deps; node_modules/karma/bin/karma start --single-run karma.config.js; fi"
+  - sh -c "if [ '$JSTESTS' = '1' ]; then npm install --deps; node_modules/karma/bin/karma start karma.config.js --single-run; fi"
   - cd ../
 
 matrix:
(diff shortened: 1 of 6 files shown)


PVince81:
@nickvergessen JS tests can run and pass: https://travis-ci.org/owncloud/activity/jobs/75584373#L406
I think this is ready to merge.

.travis.yml
```diff
@@ -47,6 +47,11 @@ script:
   - sh -c "if [ '$CODECHECK' != '1' -a '$CODECHECK' != '2' -a '$TRAVIS_PHP_VERSION' != 'hhvm' ]; then php ocular.phar code-coverage:upload --format=php-clover clover.xml; fi"
```
nickvergessen: add ` -a '$JSTESTS' != '1'` to all `'$CODECHECK' != '1' -a '$CODECHECK' != '2'`, so we only run either JS or php unit tests so they fail individually :wink:

PVince81: ah yes, good point.

PVince81:
@nickvergessen CI passed, merge ?

nickvergessen:
Tested and can run local :+1: 
Thanks!

PR #364, opened 2015-08-14
Files activity tab eventdispatcher

DeepDiver1975:
https://github.com/owncloud/activity/pull/358 with eventdispatcher

@nickvergessen @PVince81


PVince81:
Obsolete. I used a similar approach on the other PR but it happens in FilesHooksStatic::register() instead: https://github.com/owncloud/activity/pull/358/files#diff-92ecfcb2b9843406b96daa16bb97d068R43

PR #368, opened 2015-08-20
Du not show duplicate preview icons when grouping activities

### LINKED ISSUE #272: When multiple files/folders are grouped into one message, show multiple icons/thumbnails

nickvergessen:
@jancborchardt @PVince81 

Fix #272


PVince81:
:+1:

PR #390, opened 2015-09-14
Avoid possible wrong translations of the also confusing time information

### LINKED ISSUE #387: Text of activity notification mail doesn't fit to send cycles
e.g. mail contains following statement in body:
You are receiving this email because in the last day the following things happened at xxx

If you send out notifications hourly it should be:
You are receiving this email because in the last hour the following things happened at xxx

for weekly settings you better say:
You are receiving this email because in the last week the following things happened at xxx

Another solution would be to be more general:
You are receiving this email because ~~in the last day~~ the following things happened at

nickvergessen:
Fix #387

### DIFF
--- a/lib/mailqueuehandler.php
+++ b/lib/mailqueuehandler.php
@@ -197,24 +197,6 @@ class MailQueueHandler {
 	}
 
 	/**
-	 * Get the selected setting for a time frame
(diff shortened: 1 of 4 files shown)


karlitschek:
looks good :+1:

PR #398, opened 2015-09-18
No reshare activities for owner when they share themselves

### LINKED ISSUE #397: Strange activity message flood
capture du 2015-09-17 19-31-04

I get one of each of these messages _every time_ I share via link, or just »allow editing«. There should only be one activity of course, and the technical id there is wrong. … @nickvergessen

nickvergessen:
Fix #397 

@jancborchardt


karlitschek:
looks good :+1:

jancborchardt:
Yeah, thanks!

PR #402, opened 2015-10-02
Set the formatting object so Extensions can adjust

### LINKED ISSUE #395: Reduce amount of text in activity sidebar, don’t show filename on every change
Currently in every single entry the filename is mentioned. This is way too much clutter in the sidebar where we don’t have much space and info should not be redundant anyway.
(Ignore the missing share icons please, that’s caused by my adblocker.)
capture du 2015-09-17 14-19-49
Here’s some better wording. Of course we can only use it here in the file-specific view, not in the Activity app.
capture du 2015-09-17 14-19-15

@owncloud/designers @nickvergessen

nickvergessen:
- [x] Requires https://github.com/owncloud/core/pull/19522

Fix #395 

@MorrisJobke @PVince81 

---

Unit tests fail because of wrong core branch, CI branch at https://github.com/owncloud/activity/pull/403

### DIFF
--- a/lib/grouphelper.php
+++ b/lib/grouphelper.php
@@ -185,12 +185,14 @@ class GroupHelper
 
 		$return = array();
 		foreach ($this->activities as $activity) {
+			$this->activityManager->setFormattingObject($activity['object_type'], $activity['object_id']);
 			$activity = $this->dataHelper->formatStrings($activity, 'subject');
 			$activity = $this->dataHelper->formatStrings($activity, 'message');
 
 			$activity['typeicon'] = $this->activityManager->getTypeIcon($activity['type']);
 			$return[] = $activity;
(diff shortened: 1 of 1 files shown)


nickvergessen:
Unit tests fail because of wrong core branch, CI branch at https://github.com/owncloud/activity/pull/403

PVince81:
It still says "You changed welcome.txt" in the sidebar. Is it intended ?
The sharing message however doesn't contain the file name any more.

nickvergessen:
Yeah, I didn't change the self file actions, because "Changed", "Created" and so on dont make a lot of sense...

PVince81:
Works :+1:

MorrisJobke:
Works :+1:

PR #408, opened 2015-10-08
Use l10n factory instead of non-existing method

nickvergessen:
@MorrisJobke 

Fixes broken master


MorrisJobke:
Tested and works :+1:

PR #424, opened 2015-10-26
Add an index to cover the expire SQL query

### LINKED ISSUE #423: Expiring takes a long time
I'm running 8.1.3 and have about 10,000,000 rows in the activity table. It takes about 2 minutes to run the delete from oc_activity where `timestamp`<? query in the expiry job even if there are no rows that need to be deleted. An explain on that query shows that there are no indexes being used. However I see timestamp is part of 3 other indexes.

However if the cron job doesn't run for a few days for whatever reason, there are so many records that do need to be deleted that this query takes forever.

Should timestamp be in the primary key or be it's own index to speed up this delete operation? Perhaps someone with more MySQL experience could comment?

nickvergessen:
@karlitschek we could backport this easily.
The issue was reported for 8.1.3
Should we backport it? And if yes, only to 8.2 or also 8.1?

Fix #423 

@scolebrook

### DIFF
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -104,6 +104,14 @@
 			</field>
 
 			<index>
+				<name>activity_time</name>
+				<field>
+					<name>timestamp</name>
+					<sorting>descending</sorting>
+				</field>
+			</index>
+
+			<index>
 				<name>activity_user_time</name>
 				<field>
 					<name>affecteduser</name>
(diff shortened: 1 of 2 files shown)


karlitschek:
This makes sense. But let's not do schema changes in patch releases. So no backport here please.

DeepDiver1975:
Okay - generating this index might take a hell of time - thinking about big installations with millions of activity entries. Just saying ....

nickvergessen:
Okay, so we keep it for 9.0 only and people asking for it can create it manually.
Works for me

scolebrook:
@DeepDiver1975 @karlitschek @nickvergessen I agree. We will do this manually when we can be offline for an hour or so for the alter to complete.

scolebrook:
@nickvergessen And thanks for the speedy fix.

PR #425, opened 2015-10-26
Fix the back and forward buttons on activitiy filters

nickvergessen:
Thanks for teaching @PVince81 

Mind reviewing?


PVince81:
:+1:

PR #432, opened 2015-11-10
use new checkbox style for in-app settings

jancborchardt:
Please review @nickvergessen @Henni @owncloud/designers 

We should also backport this since the new checkboxes should be used for 8.2 everywhere. @karlitschek

### DIFF
--- a/templates/stream.app.navigation.php
+++ b/templates/stream.app.navigation.php
@@ -27,7 +27,7 @@
 		</div>
 
 		<div id="app-settings-content">
-			<input type="checkbox"<?php if ($_['rssLink']): ?> checked="checked"<?php endif; ?> id="enable_rss" />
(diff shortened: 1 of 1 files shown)


MorrisJobke:
Tested and works :+1:

nickvergessen:
:+1:

nickvergessen:
Backport shouldn't hurt, but is also not that strongly visible, so no need to kill kittens for bringing it in 8.2.1, can simply do it after that.

karlitschek:
please backport :+1:

jancborchardt:
Backported 1a75902

nickvergessen:
Please always use branches for backports in the future

jancborchardt:
@nickvergessen ok, sorry – thought we only use that workflow for core already. Will do in the future.

Issue #405, opened 2015-10-08
API: No way to retrieve events later than a certain point 


dragotin:
The current API only takes the parameters `page number` and `number of entries` to influence the subset of entries returned.

For a client that is not suitable. We need an API that provides all new entries that came after a certain point of time. That can either be done through ID or timestamp. The client needs a way to specify a number of events to be returned at a time (already there), but also a way to query how many new entries are new after a certain point. 

This blocks https://github.com/owncloud/client/issues/3732, https://github.com/owncloud/client/issues/3728

DeepDiver1975:
> For a client that is not suitable. We need an API that provides all new entries that came after a certain point of time.

Reading this I assume you intend to store the activities locally in the clients and want to check in a given interval for new activities - right?

So basically this would be an activity sync mechanism.

nickvergessen:
Don't really care too much, which route we go. Either OCS or [WebDAV](https://github.com/owncloud/activity/pull/206)

OCS has the pro that it exists already, con it needs to be adjusted a lot to allow starting by timestamp and more, so we either break the existing endpoint or need to make a new one, so the already-existing benefit is gone.

Anyone with power please make the call.

DeepDiver1975:
> Don't really care too much, which route we go. Either OCS or WebDAV

Protocol is to be choosen at a later stage - for now I want to understand the client's expectations and it's operations.

@dragotin

dragotin:
I was thinking about this:
- Client maintains a local storage of the events to be able to display them quickly and also in offline mode
- From that local storage, client can find out the latest event it knows, ie. the highest timestamp, or maybe better id.
- to fetch new events from the server, it sends the last known highest timestamp along and the server returns the new events that came after that.
- To not get completely flooded client sends a maximum of events per call, and server replies accordingly with that amount maximum. If there are more that has to be indicated somehow.

But maybe I misunderstood the API completely?

nickvergessen:
Well the current API is like a paged search. Just like you would get from a book or in the web ui.
However the local storing and offline thing makes sense to me and I also see no other way then having a "I know everything up to activity #X" approach.

karlitschek:
@nickvergessen OCS please. Let's stay at the current route with the current API unless a change is absolutely necessary.

nickvergessen:
## Endpoint

1. Specify the last known ID
2. Sorts ascending
3. Header specifies the first known ID (in case the given one was pruned (time))
[...]

PR #434, opened 2015-11-20
Add a new OCS endpoint that allows to give an offset

nickvergessen:
Working further towards #405 

## Todo's
- [ ] Use the `*_prepared` in the webUI instead of the preparsed stuff
- [x] Adjust unit tests
  - [x] Controller
  - [x] Controller::get()
  - [x] Data
  - [x] Others

### Parameters
- [x] since
- [x] limit
- [x] sort (order)
- [x] object type&id

### Status codes (OCS v2)
- [x] 200 when activites are given
- [x] 304 when there are no activities
- [x] 403 when the since belongs to another user
- [x] 403 when the user is not logged in
- [x] 404 when the filter is unknown

### Headers
- [x] Send `Link` when there are further results
- [x] Send `X-Activity-First-Known` when the given since does not exist anymore
- [x] Send `X-Activity-Last-Given` since parameter for the next call

### Element
- [x] Replace timestamp with ISO date
- [x] Change parsing to something client neutral

### Usage
- [x] Replace existing endpoints with calls to OCS
- [x] Replace existing calls to `Data::read()` with `Data::get()`


nickvergessen:
This is completed now @DeepDiver1975 @dragotin 
Shall I just go ahead and merge it like this, or do you want to test on it @dragotin 

Full spec is added as a markdown version as well:
https://github.com/owncloud/activity/blob/ocs-for-all/docs/ocs-endpoint-v2.md

PR #455, opened 2016-01-28
Fix OC_L10N usage

rullzer:
Now I can browse my activities again :)

CC: @nickvergessen

### DIFF
--- a/lib/navigation.php
+++ b/lib/navigation.php
@@ -23,10 +23,11 @@
 
 namespace OCA\Activity;
 
-use \OCP\Activity\IManager;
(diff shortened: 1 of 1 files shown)


nickvergessen:
Hu? That should still work? it's only deprecated, not removed?

rullzer:
Yes... but https://github.com/owncloud/activity/blob/master/appinfo/application.php#L62 returns an instance of `\OC\IL10N\L10N` not OC_L10N.

Which gave an error on my php7 install...

nickvergessen:
Ah okay

PR #462, opened 2016-02-12
Run build/license.php from core

nickvergessen:


rullzer:
What can I say.... look good... :+1: and my activities are still working

PR #463, opened 2016-02-17
Move the API back to a controller, until the client guys approved the…

nickvergessen:
… format


DeepDiver1975:
@nickvergessen is this API used at all at the moment? If not we will kill this for 9.0 - no need to ship anything which is unused.

nickvergessen:
It's used in the web UI, which is why I removed the OCS urls and moved it back to controller urls.

PR #473, opened 2016-03-09
fix message style in activity view

jancborchardt:
Make it normal size and left-align it with the avatar, so that it’s like in the sidebar in Files.

Please review @nickvergessen @owncloud/designers 

I’d also like to backport this to stable9 since this looks like a bad design glitch. @karlitschek

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -13,7 +13,7 @@
 	overflow: hidden;
 }
 .activitymessage {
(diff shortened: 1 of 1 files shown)


karlitschek:
:+1:

ChristophWurst:
:+1:

PR #486, opened 2016-03-14
Do not register CSRF token

LukasReschke:
The session is closed at this point since it may come over CLI => :bomb:

Fixes https://github.com/owncloud/core/issues/23205

cc @nickvergessen 
cc @karlitschek Backport to 9.0.1?

### DIFF
--- a/lib/mailqueuehandler.php
+++ b/lib/mailqueuehandler.php
@@ -247,7 +247,7 @@ class MailQueueHandler {
 			);
 		}
 
-		$alttext = new Template('activity', 'email.notification', '');
+		$alttext = new Template('activity', 'email.notification', '', false);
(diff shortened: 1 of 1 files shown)


nickvergessen:
Have the same PR locally.
I wonder if the template thing should check not only for ->getSession() but also see if it is closed?

nickvergessen:
https://github.com/owncloud/core/blob/349f3bf54a81af8467a62f1681bcaca172262446/lib/private/template.php#L75-L75

karlitschek:
good catch. please backport :+1:

LukasReschke:
> Have the same PR locally.
> I wonder if the template thing should check not only for ->getSession() but also see if it is closed?

Good point. Will make a PR for core's master on this. :+1:

PR #490, opened 2016-03-16
Ignore \OCP\Files\NotFoundException on chunk assembling

nickvergessen:
File not found? Sounds weird, but this happens before 9.1:
https://github.com/owncloud/core/issues/23212
Chunk assembling triggered the exact same hooks twice.
The first call however is before the file is in the database.
So when trying to get the owner, the file can not be found.
But since the second hook will come along, we simply ignore this.

### DIFF
--- a/lib/fileshooks.php
+++ b/lib/fileshooks.php
@@ -142,7 +142,21 @@ class FilesHooks {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


PVince81:
:+1:

Issue #504, opened 2016-04-11
Combine comments and activity tab in the File sidebar


nickvergessen:
- [ ] "List your own actions in the stream" should only affect file actions (add, change, delete, restore, share). Especially comments activities should be visible.
- [ ] Scrolling/show more to retrieve earlier comments: https://github.com/owncloud/activity/issues/413
- [ ] Add an JS event for comments to add the "New comment" box
- [ ] Add an JS event to allow apps to help formatting their activities (e.g. add delete/edit buttons on comments)

- [ ] Make sure the comments app still works, also when activities are disabled :exclamation:

PVince81:
@pmaier1 would move to backlog...

pmaier1:
Yes. Btw. aren't these quite identical https://github.com/owncloud/core/issues/20259 or is this just the 'activity' part of it?

PVince81:
They're similar. I don't remember the differences. Need to take some time to synthesize and/or reconsider the approach.

DeepDiver1975:
if we want a combined view we need a combined api .... one day ...

PR #505, opened 2016-04-11
"list self action" should only affect file create, change, delete and restore

nickvergessen:
- [x] Requires https://github.com/owncloud/core/pull/23915

### Steps
1. Make sure the checkbox is **not** checked
* Upload a file
* Comment on a file
* Check that you have an activity about the comment, but not about the upload

Ref #504 Step 1.


ChristophWurst:
:+1: tested and works

schiesbn:
tested... works :+1:

PR #512, opened 2016-04-19
Reduce the number of 4000chars columns

nickvergessen:
@DeepDiver1975 for the MySQL 4byte work.
Does text(4000) to clob work on all dbms?

### DIFF
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -51,7 +51,7 @@
 				<name>app</name>
 				<type>text</type>
 				<notnull>true</notnull>
-				<length>255</length>
+				<length>32</length>
 			</field>
 			<field>
 				<name>subject</name>
@@ -61,9 +61,8 @@
 			</field>
 			<field>
 				<name>subjectparams</name>
-				<type>text</type>
+				<type>clob</type>
 				<notnull>true</notnull>
-				<length>4000</length>
 			</field>
 			<field>
 				<name>message</name>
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
> Does text(4000) to clob work on all dbms?

We have to test this

nickvergessen:
Okay, will try to add one in core.

LukasReschke:
I guess @butonic won't like you for that: https://github.com/owncloud/core/issues/23895#issuecomment-208184306 😉

butonic:
:-1: find a better solution for oracle first

nickvergessen:
Just saying in the activity app we never sort, search or whatever with these columns, only select via * is done.

DeepDiver1975:
Using clob is valid from my pov - but the migration is tricky.
We also have to take into account that the activity table is hugh on some installations .....

karlitschek:
This will make the upgrade from 9.0 to 9.1 slow. I'm not saying that we can't do that but we have to be careful. 
I also agree with @DeepDiver1975 and @butonic that this needs proper testing. Experience shows that there is a lot that can go wrong in corner-cases.

nickvergessen:
So do we do this now, or one more version until people with 4byte mysql can use activity app?

nickvergessen:
Oracle is unable to convert from text to clob atm, see https://github.com/owncloud/core/issues/24897
So if we don't want people to run some commands manually for the time-being, there is no way to do that.

PR #513, opened 2016-04-19
use $().prop('checked', ...) instead of $().attr('checked', ...)

ChristophWurst:
Discovered while working on https://github.com/owncloud/core/pull/24081


nickvergessen:
:+1: works, this is set to "to develop" but I think we can merge it?

PR #515, opened 2016-04-21
Add a button to load more activities in the sidebar

nickvergessen:
Fix #413

PR #516, opened 2016-04-26
Fix displaying of the date in the sidebar

nickvergessen:
- [x] Depends on #515 

Should backport the last commit to 9.0: https://github.com/owncloud/activity/commit/d4a59fd7f850e25dc8bbedc4936bc266d63c4a90


PVince81:
Tested, works :+1: 
The date now appears.
Looks a bit weird as the date seems to be on the next row on my screen but no biggie, can be fixed separately.

PR #518, opened 2016-04-26
Allow OCA.Activity.Plugins to help rendering activities in the sidebar

nickvergessen:
Required for #504 Checkbox 3
Required for owncloud/core#24275

### DIFF
--- a/js/activitytabview.js
+++ b/js/activitytabview.js
@@ -48,6 +48,7 @@
 		},
 
 		_loading: false,
+		_plugins: [],
 
 		initialize: function() {
 			this.collection = new OCA.Activity.ActivityCollection();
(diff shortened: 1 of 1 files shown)


js/activitytabview.js
```diff
@@ -56,6 +57,8 @@
 			this.collection.on('sync', this._onEndRequest, this);
 			this.collection.on('error', this._onError, this);
 			this.collection.on('add', this._onAddModel, this);
+
+			this._plugins = OC.Plugins.getPlugins('OCA.Activity.Plugins');
```
PVince81: I'd suggest using a more specific name, maybe `RenderingPlugins` or so. In the future you might want to have several types of plugins doing different things.

If you don't, it means you'll have to detect the plugin type with a method `getType` or `type` attribute of some sorts, which is also ok.

nickvergessen: Yeah I guess that makes sense

PVince81:
Rebased and tested, works :+1:

PR #521, opened 2016-05-06
Use permalinks in activities

PVince81:
- [x] REQUIRES: permalinks in core https://github.com/owncloud/core/pull/24434
- [ ] TODO: fix unit tests
- [x] TODO: trashbin cases ? (might be a job for core...)
- [ ] TEST: all three cases, especially the one with "getInfoByPath" where we get the fileid from `$info`

Are there other cases I forgot ?

### DIFF
--- a/lib/Controller/OCSEndPoint.php
+++ b/lib/Controller/OCSEndPoint.php
@@ -1,6 +1,7 @@
 <?php
 /**
  * @author Joas Schilling <nickvergessen@owncloud.com>
+ * @author Vincent Petry <pvince81@owncloud.com>
  *
  * @copyright Copyright (c) 2016, ownCloud, Inc.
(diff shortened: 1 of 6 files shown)


nickvergessen:
Well since it doesn't work for trashbin and the old way does, I'd say we shouldn't do that atm:
https://github.com/owncloud/core/pull/24434#discussion_r62339111

PVince81:
- [ ] REQUIRES: https://github.com/owncloud/core/pull/24537 for trash support

PVince81:
doesn't work, needs rework, no time

PVince81:
rebased

PVince81:
not a priority currently and there are technical challenges that will take longer to solve.
activity app concept needs to be rethought on a higher level

PR #530, opened 2016-08-01
fix 529: Avatar image in activity tab broken for "remote user"

### LINKED ISSUE #529: Avatar image in activity tab broken for "remote user"
### Steps to reproduce
1. Enable Activity App
2. Create "editable" public share
3. Upload something into it using the public link
4. As the owner, check the activities tab for the uploaded object

### Expected behaviour
The entry created by "remote user" should have either some generic or no avatar.

### Actual behaviour
owncloud_activity_remoteuser
Entry for "remote user" links to non-existing avatar image, showing ugly placeholder instead.

### Server configuration
* ownCloud 9.1
* Activities 2.3.2

### Analysis
_js/activitytabview.js:170_ _element.avatar(element.data('user'), 28)_ generates a broken link if user is empty.
The main activity view _js/script.js:254_ uses the (first character of) the display name as "icon" if a username is not available. In the case of "remote user" (hard coded in _lib/Formatter/UserFormatter.php:54_), that's the quotation mark. Whether that's the best option is open to discussion, however it's definitely better than the broken image. I'll submit a pull request to do the same for the tab view shortly.

fmkaiser:
Fix #529: Use (first character of) display name for avatar in activity tab if username is unavailable same as main in view.

### DIFF
--- a/js/activitytabview.js
+++ b/js/activitytabview.js
@@ -166,7 +166,11 @@
 		_postRenderItem: function($el) {
 			$el.find('.avatar').each(function() {
 				var element = $(this);
-				element.avatar(element.data('user'), 28);
+				if (element.data('user-display-name')) {
+					element.avatar(element.data('user'), 28, undefined, false, undefined, element.data('user-display-name'));
+				} else if (element.data('user')) {
+					element.avatar(element.data('user'), 28);
+				}
 			});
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
:+1:

PVince81:
stable9.1: https://github.com/owncloud/activity/pull/540

PR #546, opened 2016-11-22
Use migrations to adjust the database schema to be ready for use with…

DeepDiver1975:
… mysql mb4

### DIFF
--- a/appinfo/Migrations/Version20161122085340.php
+++ b/appinfo/Migrations/Version20161122085340.php
@@ -13,6 +13,8 @@ class Version20161122085340 implements ISchemaMigration {
 		if ($schema->hasTable("{$prefix}activity")) {
 			return;
 		}
+
+		// not that valid ....
 		$schemaReader = new MDB2SchemaReader(\OC::$server->getConfig(), \OC::$server->getDatabaseConnection()->getDatabasePlatform());
 		$schemaReader->loadSchemaFromFile(__DIR__ . '/../database.xml', $schema);
 	}
--- a/appinfo/Migrations/Version20161122085340.php
+++ b/appinfo/Migrations/Version20161122085340.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace OCA\activity\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use OC\DB\MDB2SchemaReader;
+use OCP\Migration\ISchemaMigration;
+
+class Version20161122085340 implements ISchemaMigration {
+
+	public function changeSchema(Schema $schema, array $options) {
+		$prefix = $options['tablePrefix'];
+		if ($schema->hasTable("{$prefix}activity")) {
+			return;
+		}
+		$schemaReader = new MDB2SchemaReader(\OC::$server->getConfig(), \OC::$server->getDatabaseConnection()->getDatabasePlatform());
+		$schemaReader->loadSchemaFromFile(__DIR__ . '/../database.xml', $schema);
+	}
+
+}
--- a/appinfo/Migrations/Version20161122092159.php
+++ b/appinfo/Migrations/Version20161122092159.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace OCA\activity\Migrations;
+
+use Doctrine\DBAL\Platforms\OraclePlatform;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Types\Type;
+use OCP\Migration\ISchemaMigration;
+
+class Version20161122092159 implements ISchemaMigration {
+
+	/**
+	 * @param Schema $schema
+	 * @param array $options
+	 */
+	public function changeSchema(Schema $schema, array $options) {
+		$prefix = $options['tablePrefix'];
+		$tableName = "{$prefix}activity";
+		$table = $schema->getTable($tableName);
+		// we only apply this step if the columns is not yet a CLOB
+		if ($table->getColumn('subjectparams')->getType() === Type::getType(Type::TEXT)) {
+			return;
+		}
+
+		if (\OC::$server->getDatabaseConnection()->getDatabasePlatform() instanceof OraclePlatform) {
+			return;
+		}
+		$table = $schema->getTable($tableName);
+		$table->changeColumn('subjectparams', ['type' => Type::getType('text')]);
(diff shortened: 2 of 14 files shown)


appinfo/Migrations/Version20161122092159.php
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public function up(Schema $schema) {
+		$prefix = $this->connection->getPrefix();
+		$tableName = "{$prefix}activity";
+		if ($this->connection->getDatabasePlatform() instanceof OraclePlatform) {
+			$tableName = $this->connection->quoteIdentifier($tableName);
+			$this->addSql("ALTER TABLE $tableName ADD (\"tmpsubjectparams\" CLOB, \"tmpmessageparams\" CLOB)");
+			$this->addSql("UPDATE $tableName SET \"tmpsubjectparams\"=\"subjectparams\", \"tmpmessageparams\"=\"messageparams\"");
+			$this->addSql("COMMIT");
+			$this->addSql("ALTER TABLE $tableName DROP COLUMN \"subjectparams\"");
```
butonic: DROP takes time. Instead of DROP rename it to a different name, eg "oldsubjectparams" and "oldmessageparams", then rename the temp colums. This keeps the gap to a minimum ... the old* colums can then be deleted after the migration ...

basically we have to decide if we want a fast migration (which requires more disk space and a subsequent cleanup) or a migration that keeps a clean state ...

I vote for fast.

appinfo/Migrations/Version20161122092159.php
```diff
@@ -0,0 +1,44 @@
...
+class Version20161122092159 extends AbstractMigration {
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema) {
+		$prefix = $this->connection->getPrefix();
+		$tableName = "{$prefix}activity";
+		if ($this->connection->getDatabasePlatform() instanceof OraclePlatform) {
+			$tableName = $this->connection->quoteIdentifier($tableName);
+			$this->addSql("ALTER TABLE $tableName ADD (\"tmpsubjectparams\" CLOB, \"tmpmessageparams\" CLOB)");
```
butonic: if you are not using `$this->connection->quoteIdentifier('tmpsubjectparams');` you might as well use `ALTER TABLE \"$tableName\"` ...

butonic: do not make these colums CLOB, then we might as well say activities is unsupported on oracle. In https://github.com/owncloud/activity/blob/master/lib/Data.php#L207 we execute a `SELECT *` which causes the automatic conversion of OCI-Lob objects into strings. In the case of the appconfig table that easily adds a millisecond per row, for ~120 rows that makes 120ms ON EVERY REQUEST. Here we have the activity table with thousands of rows and two CLOBS. The default limit is 30 ... or 50 ... I would expect 60-100ms for these queries.

- [ ] in https://github.com/owncloud/activity/blob/master/lib/Data.php#L303 we only need `affecteduser` and `timestamp`, get rid of the `*`

- [ ] alse LIKE does not work on CLOB columns: https://github.com/owncloud/activity/blob/master/appinfo/update.php#L38 in this case we might get away with it if the cleanup happens before this up script.

appinfo/Migrations/Version20161122092159.php
butonic: no, you need the space anyway because you have to duplicate the columns anyway ...

appinfo/Migrations/Version20161122092159.php
butonic: ok, CLOB might work ... performance will suck big time compared to the open source databases ... but might be acceptable.

DeepDiver1975: I guess we can remove these update steps - 1.2.2 is oc 8.1 afaik

DeepDiver1975:
ALL GREEN!

@PhilippSchaffrath @PVince81 @butonic please review and test! THX

appinfo/Migrations/Version20161122085340.php
```diff
@@ -0,0 +1,31 @@
...
+namespace OCA\activity\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use OC\DB\MDB2SchemaReader;
+
+class Version20161122085340 extends AbstractMigration {
+
+	/**
+     * @param Schema $schema
```
PhilippSchaffrath: Indentation swaps somewhat from tabs to spaces here and in the rest of this file.

PVince81:
smiley-activity


```
MariaDB [owncloud]> select * from oc_activity where file like '%☻%';
+-------------+------------+----------+--------------+-------+--------------+-------+--------------+---------------------------------------+---------+---------------+---------------+-----------------------------------------------------------+-------------+-----------+
| activity_id | timestamp  | priority | type         | user  | affecteduser | app   | subject      | subjectparams                         | message | messageparams | file          | link                                                      | object_type | object_id |
+-------------+------------+----------+--------------+-------+--------------+-------+--------------+---------------------------------------+---------+---------------+---------------+-----------------------------------------------------------+-------------+-----------+
|           8 | 1486031825 |       30 | file_created | user1 | user1        | files | created_self | [{"26":"\/test\/\u263b.txt"}]         |         | []            | /test/☻.txt   | http://localhost/owncloud/index.php/apps/files/?dir=/test | files       |        26 |
|           9 | 1486031825 |       30 | file_created | user1 | admin        | files | created_by   | [{"26":"\/test\/\u263b.txt"},"user1"] |         | []            | /test/☻.txt   | http://localhost/owncloud/index.php/apps/files/?dir=/test | files       |        26 |
|          10 | 1486031828 |       30 | file_changed | user1 | user1        | files | changed_self | [{"26":"\/test\/\u263b.txt"}]         |         | []            | /test/☻.txt   | http://localhost/owncloud/index.php/apps/files/?dir=/test | files       |        26 |
|          11 | 1486031828 |       30 | file_changed | user1 | admin        | files | changed_by   | [{"26":"\/test\/\u263b.txt"},"user1"] |         | []            | /test/☻.txt   | http://localhost/owncloud/index.php/apps/files/?dir=/test | files       |        26 |
+-------------+------------+----------+--------------+-------+--------------+-------+--------------+---------------------------------------+---------+---------------+---------------+-----------------------------------------------------------+-------------+-----------+
4 rows in set (0.00 sec)
```

👍

PR #551, opened 2016-12-06
Revert "fix 529: Avatar image in activity tab broken for "remote user""

PVince81:
Reverts non-critical fix that was merged by mistake during the code freeze.

@DeepDiver1975 please approve.

I'll resend a PR so we can get the fix into 9.1.4


DeepDiver1975:
:+1:

PR #552, opened 2016-12-06
fix 529: Avatar image in activity tab broken for "remote user"

PVince81:
This time only merge later for 9.1.4


PVince81:
Tested, works 👍

Issue #375, opened 2015-08-27
File and folder renames do not show up as activity


jancborchardt:
This should be the case though, as:

> `Username` renamed **oldfilename.txt** to **newfilename.txt**

@nickvergessen

nickvergessen:
Yeah moving is missing, it's actually not that simple. When moving over boards, it can be rename,  move, add and delete for different persons on the same action. But yeah definetly on the todo, now where we have ids

alvar-freude:
At the moment owncloud logs sometimes renamed folders and files as deleted+created. It may be good to detect all rename actions correct, but it's better to log a rename too then do nothing (when deleted+created, it will be logged).

As mentioned in owncloud/core#24176: 

In SQL we have a wonderful thing called triggers. Such things like history tables (like oc_activity) can't be wrong, when filled by triggers (except the triggers have bugs). And when everything is done right, it can't be manipulated by a broken app or unprivileged user (e.g. no write access to the history and changes in main table are logged automatically).
oc_filecache seems to miss a column "last_change_by" to log also the changing user.

I've done this for debugging (I try to detect moved files/folders), the SQL is here:


```SQL
-- 
-- Debug OwnCloud renamed files etc.
-- for all changed rows, add a history message in hist.filecache
-- 
-- It may be possible to log only name changes
-- for simplicity the function logs everything  
--
-- There is no permission management at the moment 
-- (revoke from public, grant to the admin user)
--

BEGIN;

CREATE SCHEMA hist;
CREATE TYPE hist.hist_type AS ENUM ('UPDATE', 'DELETE', 'TRUNCATE');
CREATE table hist.filecache 
       (
       LIKE oc_filecache EXCLUDING INDEXES EXCLUDING CONSTRAINTS, 
       operator hist.hist_type,
       valid_to TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now()
       ) WITH (FILLFACTOR=100);

CREATE INDEX hist_filecache_fileid ON hist.filecache(fileid) WITH (FILLFACTOR=100);   

CREATE OR REPLACE FUNCTION hist.filecache_changed() RETURNS TRIGGER AS
   $code$
      BEGIN
         INSERT INTO filecache VALUES (OLD.*, TG_OP::hist_type);
         RETURN NEW;
      END;
   $code$ 
   LANGUAGE 'plpgsql'
   SECURITY DEFINER
   SET search_path = hist, pg_temp;


[...]

PR #558, opened 2017-02-02
Activity rename

PVince81:
Fixes https://github.com/owncloud/activity/issues/375

TODOs:
- [x] create activity for rename
- [x] create activity for move in
- [x] create activity for move out
- [ ] properly format activity for main activity log: "U move file X to Y" with full paths
- [ ] format/filter activity for the sidebar activity log: "U renamed this file from X to Y" with basename only
- [ ] distinguish rename/move in display ?
- [x] add activity types for email/stream filter (personal page + code)
- [ ] check handling when renaming mount points

- [ ] add unit tests


PVince81:
- [ ] check what happens when moving in/out of external storage where a user has/has no access

- [ ] how does this fit in with the "only notify for favorite"

PVince81:
Made more changes. The activity appears in the list but the subject is empty.

I can't seem to make it format two paths instead of one (two parameters of type "file"). Needs further research... Might need rewriting some pieces of the formatting logic to make this work.

PVince81:
- [ ] REQUIRES core PR: https://github.com/owncloud/core/pull/27074

PVince81:
not a priority currently and there are technical challenges that will take longer to solve.
activity app concept needs to be rethought on a higher level

PR #565, opened 2017-04-04
[stable9.1] Translate the methodname in the table header

MariusTimmer:
We are using this app for our owncloud instance and found out that the method name in the table header is not translated. This patch fixes this bug.
Issue: https://github.com/owncloud/activity/issues/567


PVince81:
:+1:

To merge for 9.1.6 once 9.1.5 is out.

PVince81:
failure unrelated, seems bootstrap.php needs patching for php 7. cc @DeepDiver1975

PR #566, opened 2017-04-04
Translate the methodname in the table header

MariusTimmer:
We are using this app for our owncloud instance and found out that the method name in the table header is not translated. This patch fixes this bug for the current master.
We've tested this in our test instance.
Issue: https://github.com/owncloud/activity/issues/567


PVince81:
If this works then it means that the translation text already exists elsewhere, so this is a valid fix 👍

PR #571, opened 2017-04-20
Parse data attributes always as string

PVince81:
Fixes issue when a user id is a number.

Needs a bit of retesting because I touched the other attributes as well.


PVince81:
For some weird reason a user with a numeric id doesn't see activities about comments even though they exist in the DB.

Might be more bugs related to this.

With this change, regular users still see their comments. Only users with numeric id...

PVince81:
did not work out and risky

PR #578, opened 2017-06-28
Comment change

phil-davis:
to run Travis CI and see what is wrong with JSTESTS


phil-davis:
Was only to confirm that JSTESTS fail

PR #579, opened 2017-06-28
Fallback to old tests/lib if no tests/unit

phil-davis:
Make app automagically select the folder for the core unit test shared code


phil-davis:
Note: JSTESTS build matrix entry is failing already, seems some underlying problem with the infrastructure.
See #578 for an innocuous edit that has the same JSTESTS problem.

phil-davis:
No longer relevant

Issue #575, opened 2017-06-26
Display group display name instead of group id when possible


PVince81:
1. Enable the custom groups app: https://github.com/owncloud/customgroups
1. Create a custom group "Group One"
1. Share a folder "test" with "Group One"
1. Select the activity tab in the right sidebar for the folder "test"

Expected result: group name shows as "Group One"

Actual result: group name displays the internal id "customgroup_groupone"

@pmaier1 FYI

Fixing this should be a matter of an hour or two.

PVince81:
since there is the possibility of duplicate group names, also have a tooltip with the actual group id

pmaier1:
Yep, this definitely needs to be done for the next release of the activity app. @felixboehm

PVince81:
@sharidas one for you. There is a new method `getDisplayName` on the `IGroup` interface that should be used.

sharidas:
The changes are merged: https://github.com/owncloud/activity/pull/582

PR #582, opened 2017-07-11
Show displayName instead of group id

sharidas:
Show displayName instead of group id

Signed-off-by: Sujith H <sharidasan@owncloud.com>

### DIFF
--- a/lib/Formatter/GroupFormatter.php
+++ b/lib/Formatter/GroupFormatter.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2016, ownCloud, Inc.
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Activity\Formatter;
+
+use OCP\Activity\IEvent;
+use OCP\IGroupManager;
+use OCP\Util;
+
+class GroupFormatter implements IFormatter {
+
+	/** @var IGroupManager  */
+	protected $groupManager;
+
+	public function __construct(IGroupManager $groupManager) {
+		$this->groupManager = $groupManager;
+	}
+
+	/**
+	 * @param IEvent $event
+	 * @param string $parameter The parameter to be formatted
+	 * @return string The formatted parameter
+	 */
+	public function format(IEvent $event, $parameter) {
+		$group = $this->groupManager->get($parameter);
+		$displayName = $group->getDisplayName();
+		return '<parameter>' . Util::sanitizeHTML($displayName) . '</parameter>';
+	}
+}
+
--- a/lib/Parameter/Factory.php
+++ b/lib/Parameter/Factory.php
@@ -22,8 +22,9 @@
 namespace OCA\Activity\Parameter;
 
 
-use OCA\Activity\Formatter\IFormatter;
 use OCA\Activity\Formatter\BaseFormatter;
+use OCA\Activity\Formatter\GroupFormatter;
+use OCA\Activity\Formatter\IFormatter;
 use OCA\Activity\Formatter\CloudIDFormatter;
 use OCA\Activity\Formatter\FileFormatter;
 use OCA\Activity\Formatter\UserFormatter;
@@ -32,6 +33,7 @@ use OCP\Activity\IEvent;
 use OCP\Activity\IManager;
 use OCP\Contacts\IManager as IContactsManager;
 use OCP\IConfig;
+use OCP\IGroupManager;
 use OCP\IL10N;
 use OCP\IURLGenerator;
 use OCP\IUserManager;
@@ -48,6 +50,9 @@ class Factory {
 
 	/** @var IL10N */
 	protected $l;
+
+	/** @var IGroupManager  */
+	protected $groupManager;
 
 	/** @var ViewInfoCache */
 	protected $infoCache;
@@ -73,6 +78,7 @@ class Factory {
 								IContactsManager $contactsManager,
 								ViewInfoCache $infoCache,
 								IL10N $l,
+								IGroupManager $groupManager,
 								$user) {
 		$this->activityManager = $activityManager;
 		$this->userManager = $userManager;
@@ -80,6 +86,7 @@ class Factory {
 		$this->contactsManager = $contactsManager;
 		$this->infoCache = $infoCache;
 		$this->l = $l;
+		$this->groupManager = $groupManager;
 		$this->user = $user;
 	}
 
@@ -130,6 +137,8 @@ class Factory {
 			return new UserFormatter($this->userManager, $this->l);
 		} else if ($formatter === 'federated_cloud_id') {
 			return new CloudIDFormatter($this->contactsManager);
+		} else if ($formatter === 'group') {
+			return new GroupFormatter($this->groupManager);
 		} else {
 			return new BaseFormatter();
 		}
--- a/tests/Formatter/GroupFormatterTest.php
+++ b/tests/Formatter/GroupFormatterTest.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2016, ownCloud, Inc.
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Activity\Tests\Formatter;
+
+use OCA\Activity\Formatter\IFormatter;
+use OCA\Activity\Formatter\GroupFormatter;
+use OCA\Activity\Tests\TestCase;
+use OCP\IGroup;
+use OCP\IGroupManager;
+
+class GroupFormatterTest extends TestCase {
+
+	/** @var  \OCP\IGroupManager|\PHPUnit_Framework_MockObject_MockObject */
+	protected $groupManager;
+
+	/**
+	 * @param array $methods
+	 * @return IFormatter|\PHPUnit_Framework_MockObject_MockObject
+	 */
+	public function getFormatter(array $methods = []) {
+		if (empty($methods)) {
+			$this->groupManager = $this->createMock(IGroupManager::class);
+			return new GroupFormatter($this->groupManager);
+		} else {
+			return $this->getMockBuilder('OCA\Activity\Formatter\BaseFormatter')
+				->setConstructorArgs([
+				])
+				->setMethods($methods)
+				->getMock();
+		}
+	}
+
+	public function dataFormat() {
+		return [
+			['para<m>eter1', '<parameter>para&lt;m&gt;eter1</parameter>'],
+			['para<m>eter2', '<parameter>para&lt;m&gt;eter2</parameter>'],
+		];
+	}
+
+	/**
+	 * @dataProvider dataFormat
+	 *
+	 * @param string $parameter
+	 * @param string $expected
+	 */
+	public function testFormat($parameter, $expected) {
+		/** @var \OCP\Activity\IEvent|\PHPUnit_Framework_MockObject_MockObject $event */
+		$event = $this->getMockBuilder('OCP\Activity\IEvent')
+			->disableOriginalConstructor()
+			->getMock();
+
+		$formatter = $this->getFormatter();
+		$group = $this->createMock(IGroup::class);
+		$this->groupManager->expects($this->once())
+			->method('get')
+			->willReturn($group);
+		$group->expects($this->once())
+			->method('getDisplayName')
+			->willReturn($parameter);
+
+		$this->assertSame($expected, $formatter->format($event, $parameter));
+	}
+}
--- a/tests/Parameter/FactoryTest.php
+++ b/tests/Parameter/FactoryTest.php
@@ -23,6 +23,7 @@ namespace OCA\Activity\Tests\Parameter;
 
 use OCA\Activity\Parameter\Factory;
 use OCA\Activity\Tests\TestCase;
+use OCP\IGroupManager;
 
 class FactoryTest extends TestCase {
 	/** @var \OCP\Activity\IManager|\PHPUnit_Framework_MockObject_MockObject */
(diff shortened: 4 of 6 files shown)


lib/Formatter/BaseFormatter.php
```diff
@@ -31,6 +31,12 @@ class BaseFormatter implements IFormatter {
 	 * @return string The formatted parameter
 	 */
 	public function format(IEvent $event, $parameter) {
+		if (\OC::$server->getGroupManager()->groupExists($parameter) === true) {
```
sharidas: I am not sure if there is any case where parameter would be passed to format here. So I thought of checking if a group name exists with data in the parameter variable. If so then try to amend the variable parameter.

PVince81: I think this is likely too deep in the lower layers. What happens if a non-group parameter is used but happens to have the same value as a group name ? (sometimes there are user ids that are the same as group ids)

Please have a look how to move this on a higher level. I expect that there is code somewhere that will read the group id and set it to a parameter. Adjust that code to pass the group display name instead, or both group id and group display name.

sharidas:
There are changes I made in the core: https://github.com/owncloud/core/pull/28387

sharidas:
Tested this by creating multiple customgroups and then tried sharing with folder. Unshared the same. Also verified this change doesn't affect the users sharing folder and unsharing.

lib/FilesHooks.php
```diff
@@ -559,6 +559,22 @@ protected function addNotificationsForUser($user, $subject, $subjectParams, $fil
...
+			 * Here we do the manipulation on subjectParam variable.
+			 * Based on the subject type convert the type from string to array.
+			 * So lets say if $subjectParams[1] has "customgroup_foo bar" and the
+			 * displayname associated with the customgroup_foo bar is say "foo bar".
+			 * Then final value of subjectParams[1] will be:
+			 * Array("customgroup_foo bar", "foo bar")
+			 */
+			if (is_string($subjectParams[1])) {
+				$displayName = $this->groupManager->get($subjectParams[1])->getDisplayName();
+				$subjectParams[1] = [$subjectParams[1],$displayName];
```
PVince81: did you check whether it's possible/easier to simply add a new subject param for the display name ?

PVince81: I also have the feeling that setting the display name like this in a sub-array doesn't make sense anyway because in your other change below you're extracting that value from the array and putting it back into a single parameter. In this case why not just pass the displayName directly without the groupId, which would have the same result.

Best would likely to have both in separate parameters and have the subject/message formatters pick the display name parameter for display, and use the group id for the tooltip.

sharidas:
I made the modification to this pr. Converted the subjectparams second index to have 'gid' and 'displayName' to include the names. And instead of making change in the BaseFormatter, I made the change in the DataHelper. Let me know if it looks ok.

tomneedham:
tests?

sharidas:
Added tests.

PVince81:
Not sure why the core change is needed ? I tested this locally without the core change and it seems to work fine.

However I observed one problem: if I rename the custom group, the name doesn't change in the activity list. However if I share with a user with display name "User One" and then rename that user's display name, it also changes in the activity list. I think we should do the same for groups.

Looking at the oc_activity table, I see this:
```
+-------------+------------+----------+--------------+-------+--------------+---------------+-------------------+-----------------------------------------------------------------------------------+---------+---------------+--------------+-------------------------------------------------------+-------------+-----------+
| activity_id | timestamp  | priority | type         | user  | affecteduser | app           | subject           | subjectparams                                                                     | message | messageparams | file         | link                                                  | object_type | object_id |
+-------------+------------+----------+--------------+-------+--------------+---------------+-------------------+-----------------------------------------------------------------------------------+---------+---------------+--------------+-------------------------------------------------------+-------------+-----------+
|           1 | 1500471034 |       30 | file_created | admin | admin        | files         | created_self      | [{"3":false}]                                                                     |         | []            |              | http://localhost/owncloud/index.php/apps/files/?dir=  | files       |         3 |
|           2 | 1500471034 |       30 | file_created | admin | admin        | files         | created_self      | [{"4":"\/welcome.txt"}]                                                           |         | []            | /welcome.txt | http://localhost/owncloud/index.php/apps/files/?dir=/ | files       |         4 |
|           3 | 1500471198 |       30 | shared       | admin | admin        | files_sharing | shared_group_self | [{"4":"\/welcome.txt"},{"gid":"customgroup_cust","displayName":"c display name"}] |         | []            | /welcome.txt | http://localhost/owncloud/index.php/apps/files/?dir=/ | files       |         4 |
|           4 | 1500471341 |       30 | shared       | admin | admin        | files_sharing | shared_user_self  | [{"4":"\/welcome.txt"},"user1"]                                                   |         | []            | /welcome.txt | http://localhost/owncloud/index.php/apps/files/?dir=/ | files       |         4 |
|           5 | 1500471342 |       30 | shared       | admin | user1        | files_sharing | shared_with_by    | [{"4":"\/welcome.txt"},"admin"]                                                   |         | []            | /welcome.txt | http://localhost/owncloud/index.php/apps/files/?dir=/ | files       |         4 |
+-------------+------------+----------+--------------+-------+--------------+---------------+-------------------+-----------------------------------------------------------------------------------+---------+---------------+--------------+-------------------------------------------------------+-------------+-----------+
```

From what I see, the user's display name is not stored in this table but is resolved at the time where the activity is rendered.

@sharidas can you have a look if it is also possible to use a similar logic for groups ?

sharidas:
Here is a glimpse of my update: 
activity

And the db query:
```
sqlite> select * from oc_activity;
1|1500534032|30|file_created|admin|admin|files|created_self|[{"8":""}]||[]||http://localhost/testing/index.php/apps/files/?dir=|files|8
2|1500534032|30|file_created|admin|admin|files|created_self|[{"9":"\/welcome.txt"}]||[]|/welcome.txt|http://localhost/testing/index.php/apps/files/?dir=/|files|9
3|1500534075|30|file_created|admin|admin|files|created_self|[{"20":"\/foo"}]||[]|/foo|http://localhost/testing/index.php/apps/files/?dir=/|files|20
4|1500534081|30|shared|admin|admin|files_sharing|shared_group_self|[{"20":"\/foo"},{"gid":"customgroup_test group1","displayName":"test group1"}]||[]|/foo|http://localhost/testing/index.php/apps/files/?dir=/foo|files|20
5|1500534084|30|shared|admin|admin|files_sharing|shared_group_self|[{"20":"\/foo"},{"gid":"customgroup_test group2","displayName":"test group2"}]||[]|/foo|http://localhost/testing/index.php/apps/files/?dir=/foo|files|20
6|1500534088|30|shared|admin|admin|files_sharing|shared_user_self|[{"20":"\/foo"},"user1"]||[]|/foo|http://localhost/testing/index.php/apps/files/?dir=/foo|files|20
7|1500534088|30|shared|admin|user1|files_sharing|shared_with_by|[{"20":"\/foo"},"admin"]||[]|/foo|http://localhost/testing/index.php/apps/files/?dir=/foo|files|20
8|1500534095|30|shared|admin|admin|files_sharing|shared_user_self|[{"20":"\/foo"},"user2"]||[]|/foo|http://localhost/testing/index.php/apps/files/?dir=/foo|files|20
9|1500534095|30|shared|admin|user2|files_sharing|shared_with_by|[{"20":"\/foo"},"admin"]||[]|/foo|http://localhost/testing/index.php/apps/files/?dir=/foo|files|20
sqlite> 

```
The group name remains ```test group2``` where as the display name is changed.
The user name is also different.

lib/DataHelper.php
```diff
@@ -179,6 +187,24 @@ public function parseParameters($parameterString) {
 		if ($parameterString[0] === '[' && substr($parameterString, -1) === ']' || $parameterString[0] === '"' && substr($parameterString, -1) === '"') {
 			// ownCloud 8.1+
 			$parameters = json_decode($parameterString, true);
+
+			/*
+			 * Remove gid from the parameters array
+			 * While displaying in the UI we want only
+			 * displayName. Convert the array to string.
+			 */
+			if((count($parameters) > 1) && is_array($parameters[1])) {
```
PVince81: where is the existing logic for the user display name ? if this here was the right location in the code I'd expect to see that logic in the same place

sharidas: While debugging I was able to see that the response code was getting ```user display-name=``` .
Here is the response:
```
{"ocs":
{
    "meta":{"status":"ok","statuscode":100,"message":null},
    "data":[
        {"activity_id":8,"type":"shared","user":"admin","affecteduser":"admin","app":"files_sharing","subject":"shared_user_self","subjectparams":[{"value":{"20":"\/foo"},"type":"file"},{"value":[{"value":"user2","type":"username"}],"type":"collection"}],"message":"","messageparams":[],"link":"http:\/\/localhost\/testing\/index.php\/apps\/files\/?dir=\/foo","object_type":"files","object_id":20,"object_name":"\/foo","activity_ids":[8,6],"files":{"20":"\/foo"},"subject_prepared":"Shared with \u003Ccollection\u003E\u003Cuser display-name=\"user2 testing\"\u003Euser2\u003C\/user\u003E\u003C\/collection\u003E","message_prepared":"","typeicon":"icon-share","datetime":"2017-07-20T07:01:35+00:00"},
        
        {"activity_id":5,"type":"shared","user":"admin","affecteduser":"admin","app":"files_sharing","subject":"shared_group_self","subjectparams":[{"value":{"20":"\/foo"},"type":"file"},{"value":[{"value":"test group2","type":"group"}],"type":"collection"}],"message":"","messageparams":[],"link":"http:\/\/localhost\/testing\/index.php\/apps\/files\/?dir=\/foo","object_type":"files","object_id":20,"object_name":"\/foo","activity_ids":[5,4],"files":{"20":"\/foo"},"subject_prepared":"Shared with group \u003Ccollection\u003E\u003Cparameter\u003Etest group2\u003C\/parameter\u003E\u003C\/collection\u003E","message_prepared":"","typeicon":"icon-share","datetime":"2017-07-20T07:01:24+00:00"},
        
        {"activity_id":3,"type":"file_created","user":"admin","affecteduser":"admin","app":"files","subject":"created_self","subjectparams":[{"value":{"20":"\/foo"},"type":"file"}],"message":"","messageparams":[],"link":"http:\/\/localhost\/testing\/index.php\/apps\/files\/?dir=\/","object_type":"files","object_id":20,"object_name":"\/foo","subject_prepared":"You created \u003Cfile link=\"http:\/\/localhost\/testing\/index.php\/apps\/files\/?dir=\/foo\" id=\"20\"\u003Efoo\u003C\/file\u003E","message_prepared":"","typeicon":"icon-add-color","datetime":"2017-07-20T07:01:15+00:00"}]
}
}
```
I formatted the response a bit.
Hence figured out that it's coming from https://github.com/owncloud/activity/blob/master/lib/Formatter/UserFormatter.php#L58. Since the group data is formatted by https://github.com/owncloud/activity/blob/master/lib/Formatter/BaseFormatter.php#L34. I thought that one of the possible solution would be to have in DataHelper.

PVince81: can we have a GroupFormatter then to follow the same pattern ?

sharidas:
So the conversion of BaseFormatter to GroupFormatter is made.

lib/Formatter/GroupFormatter.php
```diff
@@ -21,16 +22,25 @@
...
 namespace OCA\Activity\Formatter;
 
+use OC\Group\Manager;
 use OCP\Activity\IEvent;
 use OCP\Util;
 
-class BaseFormatter implements IFormatter {
+class GroupFormatter implements IFormatter {
+
+	protected $groupManager;
```
PVince81: please add PHPDoc, usually looks like:
```
/**
 * @var IGroupManager
 */
```

lib/Formatter/GroupFormatter.php
```diff
@@ -21,16 +22,25 @@
...
+	public function __construct(Manager $groupManager) {
```
PVince81: replace `Manager` with `IGroupManager`, always use the interface when possible

lib/Parameter/Factory.php
```diff
@@ -131,7 +131,7 @@ protected function getFormatter($formatter) {
 		} else if ($formatter === 'federated_cloud_id') {
 			return new CloudIDFormatter($this->contactsManager);
 		} else {
-			return new BaseFormatter();
+			return new GroupFormatter(\OC::$server->getGroupManager());
```
PVince81: please receive the `IGroupManager` in the constructor of this class

tests/Formatter/GroupFormatterTest.php
```diff
@@ -22,18 +22,22 @@
 namespace OCA\Activity\Tests\Formatter;
 
 use OCA\Activity\Formatter\IFormatter;
-use OCA\Activity\Formatter\BaseFormatter;
```
PVince81: did you delete `BaseFormatterTest` completely ?

PVince81:
Code looks much better, needs a few tweaks.

tests/Formatter/GroupFormatterTest.php
sharidas: Yes I did.

sharidas:
Apart from this failure: https://travis-ci.org/owncloud/activity/jobs/255968620 , I did addressed the changes suggested in the review.

tests/Formatter/GroupFormatterTest.php
PVince81: because it's unused ?

tests/Parameter/FactoryTest.php
```diff
@@ -197,7 +204,7 @@ public function testCreateCollection() {
 
 	public function dataGetFormatter() {
 		return [
-			['base', 'OCA\Activity\Formatter\BaseFormatter'],
+			['base', 'OCA\Activity\Formatter\GroupFormatter'],
```
PVince81: shouldn't it be "group" instead of "base" ?

I have the feeling that we shouldn't delete the BaseFormatter

sharidas: I can bring back the BaseFormatter. So from the comment I get a feel if formatter is not a group or any other formatter , then https://github.com/owncloud/activity/blob/master/lib/Parameter/Factory.php#L134 should be there to take care.

sharidas:
The BaseFormatter and BaseFormatterTest are retained.

tests/Formatter/GroupFormatterTest.php
```diff
@@ -0,0 +1,80 @@
...
+	 * @param string $expected
+	 */
+	public function testFormat($parameter, $expected) {
+		/** @var \OCP\Activity\IEvent|\PHPUnit_Framework_MockObject_MockObject $event */
+		$event = $this->getMockBuilder('OCP\Activity\IEvent')
+			->disableOriginalConstructor()
+			->getMock();
+
+		$formatter = $this->getFormatter();
+		$group = $this->createMock('OC\Group\Group');
```
PVince81: mock the interface instead ? `IGroup::class`

tests/Parameter/FactoryTest.php
```diff
@@ -76,6 +79,8 @@ protected function setUp() {
 		$this->l = $this->getMockBuilder('OCP\IL10N')
 			->disableOriginalConstructor()
 			->getMock();
+
+		$this->groupManager = $this->createMock('OCP\IGroupManager');
```
PVince81: use `IGroupManager::class` here without quotes which is better. (note: doesn't work for OC < 9.1 and beyond due to old PHP versions)

sharidas:
https://github.com/owncloud/activity/pull/582/files#diff-150498e0cdc623f3b34ba087ce23b38bR140 is the location where getFormatter method gets value as "group" for var $formatter and the GroupFormatter is returned. @PVince81 Let me know if this is what you were asking?

PVince81:
@sharidas yes, but who sets the value of `$formatter` to "group" in the first place to trigger this logic ?

sharidas:
This is the call which sets https://github.com/owncloud/activity/blob/master/lib/DataHelper.php#L153 "group". This call is made by https://github.com/owncloud/activity/blob/master/lib/GroupHelper.php#L96.
The get method https://github.com/owncloud/activity/blob/master/lib/DataHelper.php#L150 will result in calling https://github.com/owncloud/activity/blob/master/lib/Parameter/Factory.php#L107.

To be more precise, https://github.com/owncloud/core/blob/master/lib/private/Activity/Manager.php#L369 is the array which has the data "group", "file" etc.

PVince81:
@sharidas thanks for the pointers. I still don't see who sets the parameter type to "group" ? Was that part of the old code ? At least I didn't see anything in your PR so was it there already but falling back to the "base" formatter until you introduced the new one ?

sharidas:
@PVince81 Yah, that was part of the old code. In my PR I don't set "group". Before applying my patch or considering my PR, it always falls back to "base" formatter. Yes you are right.

sharidas:
The issue associated with the PR: https://github.com/owncloud/activity/issues/575

PR #583, opened 2017-07-18
Use big integer activity ID

tomneedham:
So we don't overflow here. Needs testing + maybe more insight / thought?


tomneedham:
Closed in favour of https://github.com/owncloud/activity/pull/584

PR #599, opened 2017-09-14
Change summary and description

pmaier1:


appinfo/info.xml
```diff
@@ -2,15 +2,16 @@
...
-This sets the type of activity to record, as well as whether to the user sees their own activities, whether these are only available online, and whether they get an email digest on a regular basis.
+The ownCloud Activity application enables users to not only get a summarized overview of all file and folder events in their ownCloud, but also to receive notifications for such via email. The user can configure their individual Activity preferences in their personal settings and can decide in detail which file or folder actions should be listed in the Activity stream (accessible via the app launcher) and also for which file or folder actions the users wants to receive email notifications.
+The bulk email notifications can either be sent out hourly, daily or weekly to fit the specific needs of the individual user. 
 
-More information is available in the Activity documentation.
+From creation of new files or folders, to file or folder changes, moves, renames, updates, restores from trash bin, sharing activities, comments, tags and downloads from public share links - the ownCloud Activity app gathers all file or folder related actions in one place for the user to review.
```
PVince81: not renames... see https://github.com/owncloud/activity/issues/375

appinfo/info.xml
```diff
@@ -2,15 +2,16 @@
...
+For users with lot's of activity it is possible to limit the Activity stream to 'Favorites' in order to avoid noise. Furthermore the application provides filters to give users the means to maintain overview by reducing entries to relevant information.
```
PVince81: lot's => lots

appinfo/info.xml
```diff
@@ -2,15 +2,16 @@
...
+
+And there you have it - a complete overview of all file and folder activities in your ownCloud with the additional ability to receive activity notifications via email in a time interval of your choice.
+Never again miss an important event related to content in ownCloud and always be up-to-date on all activities of your files and folders.
```
PVince81: The thing is, the activity app has some expiration logic where it deletes older entries. If you setup email then you'll likely never miss anything. But you can't rely on the activity list itself.

pmaier1: Good point but as this is quite superficial I leave it as is.

appinfo/info.xml
pmaier1: corrected - thanks

appinfo/info.xml
pmaier1: corrected - thanks

PR #610, opened 2018-02-15
Chunk deleting of rows from the activity table

tomneedham:
Fixes for clustered environments, which have a limit on rows and size of data affected by each transaction due to replication concerns.

### DIFF
--- a/lib/Data.php
+++ b/lib/Data.php
@@ -23,6 +23,7 @@
 
 namespace OCA\Activity;
 
+use Doctrine\DBAL\Platforms\MySqlPlatform;
 use OCA\Activity\Exception\InvalidFilterException;
 use OCP\Activity\IEvent;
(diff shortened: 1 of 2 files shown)


tomneedham:
Need to test this manually as well - WIP

tomneedham:
So it turns out postgres, sqlite and oracle all don't like limits in their delete statements.

GrendelWWU:
Hello Tom. 
I really like that. Can I apply this patch to 10.0.4?

tomneedham:
This will apply on top of v2.3.6 and will appear in the marketplace pending QA

madsi1m:
Hi can this please be added to stable10?

DeepDiver1975:
Grab version  2.3.7 from the marketplace - this holds this change

PR #604, opened 2017-12-18
Dont try to delete items when no users in context

tomneedham:

PR #612, opened 2018-03-14
[stable10] fix delete when no users in context

tomneedham:
backport of https://github.com/owncloud/activity/pull/604 to stable10


PVince81:
activity app is in marketplace so no need for backports any more, closing

PR #620, opened 2018-05-24
Add html template to activity email notification

VicDeo:
Addresses https://github.com/owncloud/enterprise/issues/2396

To test:

0. Specify in user email and tick activity mail notifications at the personal page
1. create/delete/share some files or folders
2. run activity cron job


michaelstingl:
@VicDeo What do you mean by `run activity cron job`? How? I totally struggle triggering the SMTP delivery. mailhog is properly configured - I see the testmail for example. Can I modify something in the db to send mail faster? `oc_jobs` , `oc_jobs` or `oc_activity_mq` ?

VicDeo:
I mean `sudo -u www-data php cron.php` but an hour should pass if  **Send emails:**  for Activity is set to **hourly**

VicDeo:
text
text

html
html

michaelstingl:
Yeah, looks HTML-ified now, and picks up the installed theme:

mailhog

@VicDeo Could you verify, does it break the text-only activity template in the `templateeditor`?

VicDeo:
@michaelstingl it shouldn't. 
OTOH when a message has both HTML and text parts - most clients show HTML part by default.

templates/html.notification.php
```diff
@@ -0,0 +1,59 @@
...
+								$l->t('You are receiving this email because the following things happened at %s', [$_['owncloud_installation']])
+							) ?>
+						</p>
+						<ul>
+							<?php foreach ($_['activities'] as $activityData) { ?>
+							<li>
+								<?php print_unescaped($l->t('%1$s - %2$s', $activityData)); ?>
+							</li>
+								<?php if ($_['skippedCount']) { ?>
+									<li>
```
PVince81: this is shifted too far

VicDeo: thanks, fixed

VicDeo:
Rebased as `Merge` button was grayed.

PVince81:
@VicDeo please check whether unit tests can be added for that bit (sorry, I just remembered now)

VicDeo:
@PVince81 yep, I can add an assertion that message contains HTML part  to the existing test

VicDeo:
@PVince81 added an expectation to the `testSendEmailToUser`

Issue #613, opened 2018-03-21
Wrong share messages in the Activities


cdamken:
I saw in the activities from the sync-client and the Activity app:

**You shared oC-Install/index.html with Christian XXXX, Christian XXXX and Dmitry YYYY**

I only shared with Christian and I shared with Dmitry a different File (in different Time)

I can see in the Logs only this related line:

```
{"reqId":"H85WNLSMht5y6TQJZJgC","level":1,"time":"2018-03-20T15:41:32+01:00","remoteAddr":"87.148.35.15","user":"carlos","app":"admin_audit","method":"POST","url":"\/ocs\/v2.php\/apps\/files_sharing\/api\/v1\/shares?format=json","message":"user <carlos>, displayName <Carlos Damken> shared file \/carlos\/files\/oC-Install\/index.html with the user chris, permissions:  READ  UPDATE  SHARE, owner: carlos [CLIENT_IP: 87.148.35.15] [USER_AGENT: Mozilla\/5.0 (Macintosh; Intel Mac OS X 10_13_3) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/65.0.3325.162 Safari\/537.36]"}
```

I never shared this file with Dmitry, what can be done? where can I search for more information?

PVince81:
### Steps:

1. Create two users "user1" and "user2"
2. Create two folders "one" and "two"
3. Share "one" with "user1"
4. Share "two" with "user2"
5. Go to activity page

### Expected result
Two entries:
- "You shared 'two' with 'user2'"
- "You shared 'one' with 'user1'"

### Actual result
A single wrong entry: "You shared 'two' with 'user2'  and 'user1'"

### Versions
OC v10.0.8 + activity v2.3.6 (also happening in v2.3.4)
Also observed on latest core stable10 branch and activity master branch.

PVince81:
The "oc_activity" table contains separate entries for both actions
The OCS API delivers a grouped version of these entries.

Seems something is missing in the GroupHelper class that decides on what fields to group on.

PVince81:
PR here: https://github.com/owncloud/activity/pull/624

PVince81:
Keeping open to discuss acceptance tests @patrickjahns

PVince81:
okay, mixed up with other ticket

paurakhsharma:
Acceptance test verifying the fix at: #673

PR #624, opened 2018-06-05
Fix grouping for share messages

PVince81:
Fixes https://github.com/owncloud/activity/issues/613

Now if you share as follows:

1. share "one" with "user1"
2. share "two" with "user2"
3. share "three" with user2"

You get:
- "You shared two and three with user2"
- "You shared one with user1"

However my logic seems to have broken other grouping:
if you create three folders "x", "y", "z" they appear as separate messages.

It seems there is some confusion in the GroupHelper about what the group key and grouping parameter is supposed to be. From my understanding it should be pointing at the parameter to be grouped on, the one that will not be repeated and everything else is aggregated. This is my current implementation.

However for the three separate folder creations, the grouping parameter points at folder name (single parameter). I need to dig deeper to better understand what was intended but have the feeling that the original code is also a bit mixed up in the semantics of the group parameter.

It might be possible to use a simpler solution that doesn't break the second scenario.

Next up:
- [ ] investigate semantics of the grouping parameter (the one selected to group on)

### DIFF
--- a/lib/GroupHelper.php
+++ b/lib/GroupHelper.php
@@ -109,20 +109,30 @@ class GroupHelper {
 			abs($activity['timestamp'] - $this->groupTime) < (3 * 24 * 60 * 60)
 			&& (!isset($this->openGroup['activity_ids']) || sizeof($this->openGroup['activity_ids']) <= 5)
 		) {
-			$parameter = $this->getGroupParameter($activity);
-			if ($parameter !== false) {
-				/** @var IParameter $parameterInstance */
-				$parameterInstance = $this->openGroup['subjectparams_array'][$parameter];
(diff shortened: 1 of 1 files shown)


PVince81:
I gave this another thought and realized that I'm trying to group on multiple fields at the same time, which is not possible. It would require the code to be able to preemptively find out what activity fields are repeated and dynamically decide whether to group on the file name or the user to be able to produce messages like:
- "You shared a and b and c with user1"
- "You shared a with user1 and user2"
- and potentially also: "You shared a and b and c with user1 and user2".

The current code isn't built to support this, so I'm going to limit it to the most common use case which is likely sharing a single thing with multiple users.

PVince81:
Correct approach here: https://github.com/owncloud/activity/pull/625

PR #631, opened 2018-07-02
Fixes #630

### LINKED ISSUE #630: Thumbnails of images are not shown in activity menu
**Safari 11**
screen shot 2018-06-29 at 12 54 25
**Chrome/Firefox**
screen shot 2018-06-29 at 12 58 33

oC 10.0.9.RC2 and activity app branch ```release-2.3.7```

DeepDiver1975:


tests/Controller/OCSEndPointTest.php
```diff
@@ -564,7 +564,7 @@ public function dataGetPreview() {
 		return [
 			['author', 42, '/path', '/currentPath', true, true, false, '/preview/dir', true],
 			['author', 42, '/file.txt', '/currentFile.txt', false, true, false, '/preview/mpeg', true],
-			['author', 42, '/file.txt', '/currentFile.txt', false, true, true, '/preview/currentFile.txt', false],
+			['author', 42, '/file.txt', '/currentFile.txt', false, true, true, '/dav/files//file.txt?x=150&y=150', false],
```
PVince81: this should actually be "remote.php/dav" but the left part is missing because you didn't mock "linkTo" for remote.php

DeepDiver1975: quite valid - but there is no difference if I return remote.php from linkTo or not.
there is no functional difference with respect to the tested code.

if I return 'TheQuickBrownFox' in the mocked method the expected result will hold 'TheQuickBrownFox'

pointless from my pov

PVince81: there is one small benefit: it enforces that you actually didn't forget to call "linkTo" because if you do, the link will be wrong

PVince81:
seems codecov isn't running for the old OC version so it thinks the code is not covered

do we need to add a test running for OC < 10.0.9 ?

DeepDiver1975:
codecov can be addressed laster -> merge and ship

PVince81:
no backports needed for activity

Issue #637, opened 2018-07-17
Jquery 2 Migration


felixheidecke:
# jQuery Core 1.9 Upgrade Guide

Changes of Note in jQuery 1.9

See the full [upgrade guide at jquery.com](http://jquery.com/upgrade-guide/1.9/)

## Minor

`.click()`

- [x] js/script.js
- [x] js/settings.js

## Neglectable

`.ajax()`

- [x] js/script.js

`.data()`

- [x] js/activitytabview.js
- [x] js/script.js

not affected by changes.
however here's a PR to get rid of these later: https://github.com/owncloud/activity/pull/571

`.attr()`

- [x] js/script.js
- [x] js/settings.js

`.prop()`

- [x] js/settings.js

`.before()`

- [x] js/activitytabview.js

PVince81:
PR here for the click adjustment: https://github.com/owncloud/activity/pull/638

PVince81:
I have verified that the activity app still works **without** modifications in regards to pagination and the affected methods above.

This means that we don't need to schedule a release for the click adjustment https://github.com/owncloud/activity/pull/638 and can just let it be bundled in whatever future release will come.

Let's close this ticket once the PR is merged.

PR #638, opened 2018-08-01
Prepare for jquery 2 update

PVince81:
Replace .click() with .on('click')

Fixes https://github.com/owncloud/activity/issues/637

To test:
- [x] clicking on "more activites" link still works (pagination)
- [x] pagination still works (`before()` call will insert ittems)
- [x] clicking on column header in checkbox settings still works

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -257,7 +257,7 @@ $(function(){
 				}
 			});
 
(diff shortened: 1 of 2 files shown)


PVince81:
Retested pagination in both the activity view and in the activity sidebar tab, all works fine.

@felixheidecke please review

Issue #652, opened 2018-10-13
Missing migrations when upgrading from 8.2.11 to 10.0.10


PVince81:
## Steps to reproduce
1. Setup OC 8.2.11
1. Upgrade to 10.0.10
1. Check database
1. Setup OC 10.0.10 from scratch
1. Check database

### Expected result
Same database schema and indices.

### Actual
Missing indices

### Reason
The reason for this bug is that the [first migration](https://github.com/owncloud/activity/blob/master/appinfo/Migrations/Version20161122085340.php#L12) is simply reading the "database.xml" file instead of recreating the whole database + indices. It skips reading the schema in case the oc_activity table already exists. This is wrong because there were many changes in database.xml between 8.2.11 and 10.0.0 (the introduction of migrations), so all these changes will be missed during a 8.2.11->10.0.10 upgrade.

#### 8.2.11 tables
```
+---------------+---------------+------+-----+---------+----------------+
| Field         | Type          | Null | Key | Default | Extra          |
+---------------+---------------+------+-----+---------+----------------+
| activity_id   | int(11)       | NO   | PRI | NULL    | auto_increment |
| timestamp     | int(11)       | NO   |     | 0       |                |
| priority      | int(11)       | NO   |     | 0       |                |
| type          | varchar(255)  | YES  |     | NULL    |                |
| user          | varchar(64)   | YES  |     | NULL    |                |
| affecteduser  | varchar(64)   | NO   | MUL | NULL    |                |
| app           | varchar(255)  | NO   |     | NULL    |                |
| subject       | varchar(255)  | NO   |     | NULL    |                |
| subjectparams | varchar(4000) | NO   |     | NULL    |                |
| message       | varchar(255)  | YES  |     | NULL    |                |
| messageparams | varchar(4000) | YES  |     | NULL    |                |
| file          | varchar(4000) | YES  |     | NULL    |                |
| link          | varchar(4000) | YES  |     | NULL    |                |
| object_type   | varchar(255)  | YES  |     | NULL    |                |
| object_id     | int(11)       | NO   |     | 0       |                |
+---------------+---------------+------+-----+---------+----------------+
15 rows in set (0.00 sec)

MariaDB [owncloudtest]> show index from oc_activity;
+-------------+------------+---------------------+--------------+--------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
[...]

PR #654, opened 2018-10-19
Add missing migrations for longtext and indices

PVince81:
- set type to "longtext" for "subjectparams" and "messageparams" by increasing length
- add index "activity_time" on column "timestamp"
- add index "activity_object" on columns "object_type" and "object_id"

Fixes https://github.com/owncloud/activity/issues/652

- ~~BUG: cannot change from text to longtext because Doctrine ignores length changes, possibly related unmerged upstream PR: https://github.com/doctrine/dbal/pull/3182, also see https://github.com/doctrine/dbal/issues/2566~~ => now resorting to direct SQL

### DIFF
--- a/appinfo/Migrations/Version20181019151118.php
+++ b/appinfo/Migrations/Version20181019151118.php
@@ -1,0 +1,58 @@
+<?php
+/**
+ * @author Vincent Petry <pvince81@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\activity\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Types\Type;
+use OCP\Migration\ISchemaMigration;
+
+/**
+ * Adds migrations that were missing when running update.
+ *
+ * - set type to "longtext" for "subjectparams" and "messageparams" by increasing length
+ *
+ */
+class Version20181019151118 implements ISchemaMigration {
+	/**
+	 * @param Schema $schema
+	 * @param array $options
+	 */
+	public function changeSchema(Schema $schema, array $options) {
+		$prefix = $options['tablePrefix'];
+		
+		$activityTable = $schema->getTable("{$prefix}activity");
+
+		if (!$activityTable->hasIndex('activity_time')) {
+			$activityTable->addIndex(
+				['timestamp'],
+				'activity_time'
+			);
+		}
+
+		if (!$activityTable->hasIndex('activity_object')) {
+			$activityTable->addIndex(
+				['object_type', 'object_id'],
+				'activity_object'
+			);
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


PVince81:
Core PR to implement the same code from upstream: https://github.com/owncloud/core/pull/33231.
This helps fix the issue but we can't automatically test this due to DBAL limitations.

PVince81:
I guess it won't be possible to make Oracle comply without gymnastics similar to https://github.com/owncloud/activity/blob/master/appinfo/Migrations/Version20170131134507.php#L18

DeepDiver1975:
> Core PR to implement the same code from upstream: [owncloud/core#33231](https://github.com/owncloud/core/pull/33231).
> This helps fix the issue but we can't automatically test this due to DBAL limitations.

for the oracle part we could leave a release note and tell people what to do ... there are only a few installations .....

DeepDiver1975:
> Core PR to implement the same code from upstream: [owncloud/core#33231](https://github.com/owncloud/core/pull/33231).
> This helps fix the issue but we can't automatically test this due to DBAL limitations.

for the oracle part we could leave a release note and tell people what to do ... there are only a few installations .....

PVince81:
@DeepDiver1975 in my local tests with core I noticed that Oracle would anyway convert the column to CLOB when asking for longtext. It looks like https://github.com/owncloud/activity/blob/master/appinfo/Migrations/Version20170131134507.php#L18 is already doing that. This means that we don't need to care about Oracle in activity for the longtext columns.

PVince81:
had a chat with @DeepDiver1975 and due to the missing DBAL feature and also the fact that this kind of migration is seldom, we'll write a migration containing direct SQL statements instead, which removes the need for the core workaround PR.

PVince81:
Seems PostgreSQL has no distinction between longtext and text, it's all "text" so we likely also do not need any change.

PVince81:
updated with plain SQL statement for MySQL.

I'll have a test with PostgreSQL just to make sure there aren't other discrepancies and to confirm that no SQL change is needed.

PVince81:
I've checked with PostgreSQL and apparently the column is already "text" and has no limit:
```
                                                                Table "public.oc_activity"
    Column     |          Type           | Collation | Nullable |                     Default                      | Storage  | Stats target | Des
cription 
---------------+-------------------------+-----------+----------+--------------------------------------------------+----------+--------------+----
---------
 activity_id   | bigint                  |           | not null | nextval('oc_activity_activity_id_seq'::regclass) | plain    |              | 
 timestamp     | integer                 |           | not null | 0                                                | plain    |              | 
 priority      | integer                 |           | not null | 0                                                | plain    |              | 
 type          | character varying(255)  |           |          | NULL::character varying                          | extended |              | 
 user          | character varying(64)   |           |          | NULL::character varying                          | extended |              | 
 affecteduser  | character varying(64)   |           | not null |                                                  | extended |              | 
 app           | character varying(255)  |           | not null |                                                  | extended |              | 
 subject       | character varying(255)  |           | not null |                                                  | extended |              | 
 subjectparams | text                    |           | not null |                                                  | extended |              | 
 message       | character varying(255)  |           |          | NULL::character varying                          | extended |              | 
 messageparams | text                    |           |          |                                                  | extended |              | 
 file          | character varying(4000) |           |          | NULL::character varying                          | extended |              | 
 link          | character varying(4000) |           |          | NULL::character varying                          | extended |              | 
 object_type   | character varying(255)  |           |          | NULL::character varying                          | extended |              | 
 object_id     | bigint                  |           | not null | 0                                                | plain    |              | 
```

so we only need to handle MySQL as already in the current PR code.

appinfo/Migrations/Version20181019151118.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * @author Vincent Petry <pvince81@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud, Inc.
```
DeepDiver1975: GmbH

appinfo/Migrations/Version20181022150134.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * @author Vincent Petry <pvince81@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud, Inc.
```
DeepDiver1975: GmbH

appinfo/Migrations/Version20181019151118.php
PVince81: I copy-pasted this file from another migration, so it's wrong there as well.

I'll adjust them all then

PVince81:
fixed license row

Issue #659, opened 2018-11-22
Move CI from Travis to drone


phil-davis:
and review ``Makefile`` etc to make the test infrastructure "standard".

phil-davis:
including putting the unit tests into ``tests/unit`` and ``phpunit.xml`` into the root folder...

PR #660, opened 2018-11-23
Move tests from travis to drone

paurakhsharma:
# Description 
Move all tests from `travis.yml` to `drone.yml`.

Related issue: #659


paurakhsharma:
@phil-davis, Please review this.

composer.json
```diff
@@ -1,10 +1,20 @@
...
+  "name": "owncloud/activity",
+  "config" : {
+    "platform": {
+      "php": "5.6.37"
     }
+  },
+  "require": {
+  },
+  "require-dev": {
+    "phpunit/phpunit": "^5.7",
```
phil-davis: See https://github.com/owncloud/password_policy/pull/166/files
We can use phpunit from core and then there is no need to put it here.

composer.json
```diff
@@ -1,10 +1,20 @@
...
+    "jakub-onderka/php-parallel-lint": "^1.0",
```
phil-davis: owncloud-coding-standard is implemented. That fails if there is a PHP syntax error. So we have been removing lint tests. I think you do not need to add that here - just leave out the old lint tests that were in Travis.

.gitignore
```diff
@@ -52,10 +52,12 @@ nbproject
 
 # Tests - auto-generated files
 /tests/coverage*
-/tests/clover.xml
-/tests/js/node_modules
-/vendor/
+/tests/unit/clover.xml
```
phil-davis: this should be ``tests/output/clover.xml``

paurakhsharma:
@DeepDiver1975 please review this. And merge this if you are happy with this.

tests/unit/ApiTest.php
```diff
@@ -25,6 +25,8 @@
 use OCA\Activity\Tests\Mock\Extension;
 use OCP\Activity\IExtension;
 
+require 'TestCase.php';
```
DeepDiver1975: this should not be necessary ...... likly namespace/path issue ....

paurakhsharma:
Drone is not running. Have to wait until drone is fixed.

phpunit.xml
```diff
@@ -0,0 +1,27 @@
...
+<phpunit bootstrap="../../tests/bootstrap.php"
+		 verbose="true"
+		 failOnWarning="true"
+		 strict="true"
+		 timeoutForSmallTests="900"
+		 timeoutForMediumTests="900"
+		 timeoutForLargeTests="900"
+>
+	<testsuite name='unit'>
+		<directory suffix='Test.php'>./tests/Unit</directory>
```
phil-davis: ``Unit`` has uppercase ``U`` here.
In other repos we made it ``unit``
So we should do the same in this repo, or we will have an annoying little difference.

phil-davis:
``tests/Unit/js/package-lock.json`` seems to have been added?
Is that an accident?

.drone.yml
```diff
@@ -0,0 +1,240 @@
...
+      - make test-php-style
+    when:
+      matrix:
+        TEST_SUITE: owncloud-coding-standard
+
+  phpunit-tests:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    environment:
+      - PHP_VERSION=${PHP_VERSION}
```
phil-davis: This line can be removed. In discussions with @patrickjahns yesterday, we realized that putting this line makes no difference and does not make anything clear...
(as we touch other repos we will need to remove this line also)

individual-it: it does make a difference if you run drone locally by `drone exec` and specify the PHP version e.g.
`PHP_VERSION=7.2 drone exec` without that line it would not pick the correct version

.drone.yml
```diff
@@ -0,0 +1,240 @@
...
+       - ORACLE_PASSWORD=oracle
+       - ORACLE_DB=${DB_NAME}
+    when:
+      matrix:
+        DB_TYPE: oci
+
+matrix:
+  include:
+    # owncloud-coding-standard
+    - PHP_VERSION: 7.2
```
DeepDiver1975: coding standard should be executed woth php 5.6

.drone.yml
```diff
@@ -0,0 +1,240 @@
...
+        TEST_SUITE: javascript
+
+  codecov:
+    image: plugins/codecov:2
+    secrets: [codecov_token]
+    pull: true
+    paths:
+      - tests/output/clover.xml
+    files:
+      - '*.xml'
```
DeepDiver1975: if full path is defined above this should not be necessary

phpunit.xml
individual-it: If we make it "**u**nit" the autoloader will not find the correct classes in the namespaces, every folder is a level in the namespace and we want to have the namespace to begin with a capital U

.drone.yml
phil-davis: Yes, ``php-cs-fixer`` is laso an effective lint/syntax checker. So run i on PHP 5.6 to make sure the code only has 5.6-valid stuff.

When we get ``phan`` happening, it will do analysis in PHP7.* and effectively becomes the lint/syntax check for PHP7.* - in case there is some old code the is valid in PHP 5.6 but is removed in PHP7.*

.drone.yml
paurakhsharma: @phil-davis, I will keep it there because we need it to run drone locally. We can remove it if required.

phpunit.xml
phil-davis: It is called ``unit`` in other repos. Is there something special in this repo?

paurakhsharma: @phil-davis  afaik, Other repo doesn't have units tests inside `tests/unit/` and use `unittest` and `bootstrap` from core. Correct me if I am wrong.

phil-davis: Yes, there is something special. ``TestCase.php``
```
namespace OCA\Activity\Tests\Unit;

abstract class TestCase extends \Test\TestCase {
```
and other tests ``use` that.

It works in a folder called ``tests/Unit`` (lowercase ``t`` at the start of ``tests``) but also making lowercase ``u`` breaks it. 

What is that all about?

phil-davis: It is in ``core`` ``lib/private/legacy/app.php`` ``registerAutoloading()``
```
		if (\defined('PHPUNIT_RUN')) {
			\OC::$composerAutoloader->addPsr4($appNamespace . '\\Tests\\', $path . '/tests/', true);
		}
```
The top-level folder can be called ``tests`` but anything in the namespace under ``Tests`` has to follow case-sensitive naming for the autoloader to work.

So we don't try to mess with this right now!

paurakhsharma: Yes, it's strange. But that is the case.

phil-davis:
@DeepDiver1975 I think your review comments have been addressed.
Please review again.

.drone.yml
patrickjahns: Please double check - as this shouldn't be required.

The `environment` key is only to inject the variable in the environment inside the container. Where does that environment need to be set inside the container to pick the right version?

individual-it: it is not needed, I was wrong see https://github.com/owncloud/files_antivirus/pull/273#discussion_r237445587

phil-davis: the line has been removed

phil-davis:
@DeepDiver1975 I think your review comments have been addressed.
Please review again.
And adjust the repo to require drone, not require travis, and I  think security/snyk should look for ``tests/Unit/js/package.json``

.drone.yml
```diff
@@ -0,0 +1,237 @@
...
+    version: ${OC_VERSION}
+    pull: true
+    db_type: ${DB_TYPE}
+    db_name: ${DB_NAME}
+    db_host: ${DB_TYPE}
+    db_username: autotest
+    db_password: owncloud
+    when:
+      matrix:
+        NEED_SERVER: true
```
patrickjahns: Why is this needed?  For any step below this variable is true

.drone.yml
```diff
@@ -0,0 +1,237 @@
...
+services:
+  owncloud:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    environment:
+      - APACHE_WEBROOT=/var/www/owncloud/
+    command: [ "/usr/local/bin/apachectl", "-e", "debug", "-D", "FOREGROUND" ]
+    when:
+      matrix:
+        NEED_SERVER: true
```
patrickjahns: Related to https://github.com/owncloud/activity/pull/660/files#r237492794 - we now start apache for every iteration which is `NEED_SERVER=true` - but none of the steps require apache

There is some missmatch in the logic currently

.drone.yml
phil-davis: coding-standard matrix entry/entries do not need NEED_SERVER, so there are some matrix entries that skip this pipeline step.
AFAIK the PHPunit needs this step to happen so that core code is there, and so that the ``phpunit`` program itself is there. The ``phpunit`` that is run is the one in core.

.drone.yml
phil-davis: yes, nowadays there are 2 different things needed:
1) NEED_THE_SERVER_CODE_TO_BE_THERE
2) NEED_SERVER_TO_BE_RUNNING

PHPunit only needs point (1)

Acceptance tests need both.

We could make 2 different names for these.

Note: this is an issue in a few app repos.

phil-davis: Possible way to do this, and naming, in https://github.com/owncloud/password_policy/pull/174/commits/856188df40941c826ae48738a99b1e4f4b199cd6

paurakhsharma: @patrickjahns @phil-davis I have made the changes. Please have a look.

.drone.yml
phil-davis: It has been update to the "new standard" of having a boolean ``NEED_CORE``

.drone.yml
phil-davis: This code has been removed because actually there are no acceptance tests yet in the activity app. So this drone ``services`` section is not even needed here.

phil-davis:
@DeepDiver1975 @patrickjahns review comments have been addressed.
Please review again.
And adjust the repo to require drone, not require travis, and I think security/snyk should look for ``tests/Unit/js/package.json`` - that will let make more green ticks for CI.

Makefile
```diff
@@ -52,38 +62,72 @@ all: $(bower_deps)
...
-$(bower_deps): $(BOWER)
-	$(BOWER) install && touch $@
+# Builds the source and appstore package
+.PHONY: dist
+dist:
+	make appstore
+
+# Builds the source package for the app store, ignores php and js tests
+.PHONY: appstore
+appstore:
```
phil-davis: Question to self: was this stuff here when I approved this PR? If so, why did I approve it? And/or why did it get added in this PR? Some accidental code pasting?

The ``Makefile``` script has no definition of ``appstore_package_name`` or ``appstore_build_directory`` so this target fails.

@paurakhsharma do you remember about this?

PR #672, opened 2018-12-13
[Not that dirty solution] Get the remote user for the federated share

VicDeo:
Depends on https://github.com/owncloud/core/pull/34122

If there is no currentUser try to resolve a remote user by dispatching an event.

activity

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -110,6 +110,19 @@ class Application extends App {
 		$container->registerService('Hooks', function (IContainer $c) {
 			/** @var \OC\Server $server */
 			$server = $c->query('ServerContainer');
+			$currentUser = $c->query('CurrentUID');
+			if ($currentUser === "") {
+				// Dirty solution to get the federated user from dav v1 server
+				// FIXME: $authBackend is a global var at https://github.com/owncloud/core/blob/master/apps/dav/appinfo/v1/publicwebdav.php#L38
+				global $authBackend;
+
+				if ($authBackend instanceof \OCA\DAV\Connector\PublicAuth) {
+					$share = $authBackend->getShare();
+					if ($share instanceof \OC\Share20\Share) {
+						$currentUser = $share->getSharedWith();
+					}
+				}
+			}
 
 			return new FilesHooks(
 				$server->getActivityManager(),
@@ -119,7 +132,7 @@ class Application extends App {
 				new View(''),
 				$server->getDatabaseConnection(),
 				$server->getURLGenerator(),
-				$c->query('CurrentUID')
+				$currentUser
 			);
 		});
 
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -41,6 +41,7 @@ use OCP\AppFramework\App;
 use OCP\AppFramework\IAppContainer;
 use OCP\IContainer;
 use OCP\Util;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 class Application extends App {
 	public function __construct(array $urlParams = []) {
@@ -112,16 +113,12 @@ class Application extends App {
 			$server = $c->query('ServerContainer');
 			$currentUser = $c->query('CurrentUID');
 			if ($currentUser === "") {
-				// Dirty solution to get the federated user from dav v1 server
-				// FIXME: $authBackend is a global var at https://github.com/owncloud/core/blob/master/apps/dav/appinfo/v1/publicwebdav.php#L38
-				global $authBackend;
-
-				if ($authBackend instanceof \OCA\DAV\Connector\PublicAuth) {
-					$share = $authBackend->getShare();
-					if ($share instanceof \OC\Share20\Share) {
-						$currentUser = $share->getSharedWith();
-					}
-				}
+				// Get the federated user from dav v1 server
+				$event = $server->getEventDispatcher()->dispatch(
+					'public.user.resolve',
+					new GenericEvent('', ['user' => ""])
+				);
+				$currentUser = $event->getArgument("user");
 			}
 
 			return new FilesHooks(


PVince81:
a bit too dirty to my taste. I wonder if we could somehow store this global thing in a new service in the server container ?

@VicDeo does this require a matching core PR ?

PVince81:
it's a nice start though that proves that it can work once we get access to this info, let's iterate a bit and see if we can make it a bit less dirty

VicDeo:
@PVince81 >  I wonder if we could somehow store this global thing in a new service in the server container ?

this will introduce a cross-app dependency 
**1st way** (dav depends on activity)  inject `$authBackend` after declaration  pointhttps://github.com/owncloud/core/blob/master/apps/dav/appinfo/v1/publicwebdav.php#L38

**2nd way** (activity depends on dav) check the global variable in the same way as it is done above in the activity Application contstructor https://github.com/owncloud/activity/blob/master/lib/AppInfo/Application.php#L48

lib/AppInfo/Application.php
```diff
@@ -110,6 +110,18 @@ public function __construct(array $urlParams = []) {
 		$container->registerService('Hooks', function (IContainer $c) {
 			/** @var \OC\Server $server */
 			$server = $c->query('ServerContainer');
+			$currentUser = $c->query('CurrentUID');
+			if ($currentUser === "") {
+				// Dirty solution to get the federated user from dav v1 server
+				global $authBackend;
```
PVince81: please link to https://github.com/owncloud/core/blob/master/apps/dav/appinfo/v1/publicwebdav.php#L38 in PHP Doc

@DeepDiver1975 is the dirtiness factor too high here for a quick fix ?

VicDeo: done

PVince81: can you also raise a ticket in core to look into ways to make it cleaner, and point here ?

PVince81: @VicDeo ping, see above

PVince81: what happens if this private var ever gets removed in core ? do we get a PHP warning or it just skips it silently ? maybe at least protect against this case

VicDeo: > if this private var ever gets removed in core -

`$authBackend` will be NULL and the code will fallback to the current `Remote User` behavior. No warnings/notices as  `global $authBackend;` declares a variable in case it hasn't been declared yet.

VicDeo:
@ownclouders rebase

PVince81:
if @DeepDiver1975 does't have any objections due to high dirtiness factor but no other alternative, I'm fine merging this

DeepDiver1975:
that's really ugly .... let me think ....

What about emitting an event via the dispatcher and let the v1.php react on it to put the fed user into the event object?

This way we do not drag dav code into activity.

VicDeo:
@DeepDiver1975 emitting event from activity app and listening to it in the dav endpoint looks reversed to me... but ok, I'll try.

PVince81:
@VicDeo or emit from DAV plugin with the info and let the activity app grab and store the auth info locally ?

VicDeo:
@PVince81 too late :)
Retested and still works.
This PR Description updated, listener part has been pushed  to the core.

lib/AppInfo/Application.php
```diff
@@ -110,6 +111,15 @@ public function __construct(array $urlParams = []) {
 		$container->registerService('Hooks', function (IContainer $c) {
 			/** @var \OC\Server $server */
 			$server = $c->query('ServerContainer');
+			$currentUser = $c->query('CurrentUID');
+			if ($currentUser === "") {
+				// Get the federated user from dav v1 server
+				$event = $server->getEventDispatcher()->dispatch(
+					'activity.user.resolve',
```
PVince81: should it be "public.user.resolve" ? the event is not specific to activity and could be used by other apps (collabora??)

PVince81: what about dav v2 (new dav) ?

PVince81:
see comments. otherwise approach is ok

lib/AppInfo/Application.php
VicDeo: @PVince81 I did a quick check and found no way to modify dav storage by token using dav v2 endpoint.
v1: `curl -X PUT -d "newContent"  -H "Authorization: Basic token" http://oc/public.php/webdav/` -works
v2: `curl -X PUT -d "newContent"   -H "Authorization: Basic token" http://oc/remote.php/dav/files/` - does not work

VicDeo: updated to 'public.user.resolve' here and in core

lib/AppInfo/Application.php
```diff
@@ -110,6 +111,15 @@ public function __construct(array $urlParams = []) {
...
 			/** @var \OC\Server $server */
 			$server = $c->query('ServerContainer');
+			$currentUser = $c->query('CurrentUID');
+			if ($currentUser === "") {
+				// Get the federated user from dav v1 server
+				$event = $server->getEventDispatcher()->dispatch(
+					'public.user.resolve',
+					new GenericEvent('', ['user' => ""])
+				);
+				$currentUser = $event->getArgument("user");
```
PVince81: can you test without the core change ? assuming `$currentUser` will then be an empty string and the activity app will work like it did before ?

if yes we will not need to update min-version and just tell people to update both core and activity to get the fix

VicDeo: @PVince81 
screenshot_20190115_145708
the first entry **fix-2444**  core branch
the second entry **master**  core branch

VicDeo: There is no hard dependency on the core version. So it's up to you to decide whether we need to bump activity min-version here.

davitol:
screen shot 2019-02-13 at 14 30 57
screen shot 2019-02-13 at 14 17 40


Now remote users appears (even when their names appear like trimmed)

PR #679, opened 2019-01-17
Enable phpcs check for acceptance tests

skshetry:
This PR introduces PHPCS check for acceptance tests. This was only in core before, and uses the same phpcs and phpcs.xml  from there.


phil-davis:
Note to note above: core PR https://github.com/owncloud/core/pull/33908 loads ``phpcs`` into core ``vendor-bin/php_codesniffer`` - we probably should wait until that gets merged (if it gets merged soon) and then sort out what to do in apps (probably similar ``vendor-bin`` stuff? Because ``phpcs`` will not be pre-loaded in the core QA tarball any more.) - let's see.

skshetry:
Already implemented by #682

Issue #677, opened 2019-01-15
write more acceptance tests


individual-it:
main framework is set up for tests (#673, #676), but a lot of cases are still untested

### WebUI
- [x] Creation of new files/folders (#680)
- [x] Update files/folders (#681)
- [x] Delete/Restore files/folders (#685)
- [x] Sharing files/folders ( ~issue: #684~ ) (#686)
- [x] Sharing files/folders from another server ( #689)
- [x] Commenting #688 
- [x] Tagging #688, #691 
- [x] Deleting tags https://github.com/owncloud/activity/pull/699
- [x] Downloading from public links ( #689)
- [ ] Changes to favourites files/folders [Blocked #543]
- [x] Limit own activities to not show up https://github.com/owncloud/activity/pull/686 
- [x] Limiting other activities types and check if it appears or not https://github.com/owncloud/activity/pull/686
- [x] Check activities added by other apps 
  - [x] Ransomware Protection https://github.com/owncloud/ransomware_protection/pull/171
  - [x] files_antivirus https://github.com/owncloud/files_antivirus/pull/297
  - [x] files_texteditor https://github.com/owncloud/files_texteditor/pull/285
  - [x] firewall https://github.com/owncloud/firewall/pull/567
- [ ] ~~Administrative logs~~

### API
- [x] there is no public API, using the private API for some general listing tests https://github.com/owncloud/activity/tree/master/tests/acceptance/features/apiActivity

skshetry:
Noticed a failure in the PR #33994. But many scenarios are failing. Not related perhaps.
```
--- Failed scenarios:
    /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:69
    /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:89
    /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:109
    /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:129
    /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:151
    /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:231
    /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:260
    /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:272
    /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:281
[...]

PR #688, opened 2019-01-28
Add test for simple tagging and commenting for activities

skshetry:
This adds a simple test for tagging and commenting for activities.

Part of #677

### DIFF
--- a/tests/acceptance/features/lib/ActivityPage.php
+++ b/tests/acceptance/features/lib/ActivityPage.php
@@ -25,7 +25,6 @@ namespace Page;
 
 use PHPUnit_Framework_Assert;
 use Behat\Mink\Session;
-use Behat\Mink\Element\NodeElement;
 
 /**
  * Activity page.
@@ -43,6 +42,8 @@ class ActivityPage extends OwncloudPage {
 	protected $activityListXpath = "//div[@class='activitysubject']";
 	protected $avatarClassXpath = "(//div[@class='activitysubject'])[%s]//div[@class='avatar']";
 	protected $fileNameFieldXpath = "(//div[@class='activitysubject'])[%s]//a[@class='filename has-tooltip']";
+	protected $messageContainerXpath = "//div[@class='messagecontainer']";
+	protected $messageTextXpath = "//div[@class='activitymessage']";
 
 	protected $activityListFilterXpath = "//a[@data-navigation='%s']";
 
@@ -115,6 +116,24 @@ class ActivityPage extends OwncloudPage {
 		);
 		$filter->click();
 		$this->waitForAjaxCallsToStartAndFinish($session);
+	}
+
+	/**
+	 * Return comment message of the given activity index
+	 *
+	 * @param string $index
+	 *
+	 * @return string|null
+	 * @throws \Exception
+	 */
+	public function getCommentMessageOfIndex($index) {
+		$messages = $this->findAll('xpath', $this->messageContainerXpath);
+		$message = $messages[$index];
+		$comment = $message->find('xpath', $this->messageTextXpath);
+		if ($comment === null) {
+			return null;
+		}
+		return $this->getTrimmedText($comment);
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


tests/acceptance/features/webUIActivity/comment.feature
```diff
@@ -0,0 +1,38 @@
...
+  I want to be able to see history of the files and folders that I have commented
+  So that I know what happened in my cloud storage
+
+  Scenario Outline: Commenting on a file/folder should be listed in the activity page
+    Given using new DAV path
+    And user "user0" has been created with default attributes
+    And user "user0" has commented with content "My first comment" on file "<filepath><filename>"
+    When user "user0" browses to the activity page
+    Then the activity number 1 should contain message "You commented on <filename>" in the activity page
+    And the comment for activity number 1 with "My first comment" message should be listed on the activity page
```
paurakhsharma: This is confusing to read. Please make this step more readable.
PS. I have no suggestion 😄

tests/acceptance/features/lib/ActivityPage.php
```diff
@@ -26,6 +26,7 @@
 use PHPUnit_Framework_Assert;
 use Behat\Mink\Session;
 use Behat\Mink\Element\NodeElement;
+use SensioLabs\Behat\PageObjectExtension\PageObject\Exception\ElementNotFoundException;
```
paurakhsharma: No, need to import this?

skshetry: fixed

tests/acceptance/features/webUIActivity/comment.feature
```diff
@@ -0,0 +1,41 @@
+@webUI @insulated @disablePreviews
+Feature: Comment files/folders activities
+  As a user
+  I want to be able to see history of the files and folders that I have commented
+  So that I know what happened in my cloud storage
+
+  Scenario Outline: Commenting on a file/folder should be listed in the activity page
+    Given using new DAV path
```
individual-it: put that in the Background, but I guess you even don't need it as your "When" steps don't use any WebDAV

tests/acceptance/features/webUIActivity/comment.feature
```diff
@@ -0,0 +1,41 @@
...
+      | 'single'quotes/simple-empty-folder/ | for-git-commit      |
+
+  Scenario Outline: Comment, and then deleting comment on a file/folder should be listed in the activity page without any comment
+    Given using new DAV path
+    And user "user0" has been created with default attributes
+    And user "user0" has commented with content "My first comment" on file "<filepath><filename>"
+    And user "user0" has deleted the last created comment
+    When user "user0" browses to the activity page
+    Then the activity number 1 should contain message "You commented on <filename>" in the activity page
+    And the activity number 1 should not contain any comment message in the activity page
```
individual-it: Is that what supposed to happen? Or is this a bug?

tests/acceptance/features/webUIActivity/comment.feature
skshetry: Comment has not been implemented for old DAV path (we debugged this the day before yesterday).

tests/acceptance/features/webUIActivity/comment.feature
skshetry: I think it should be. Whatever activity is there should be tracked by this app, and if someone deletes something, then it should protect privacy while also preserving logs.

tests/acceptance/features/webUIActivity/comment.feature
skshetry: fixed

tests/acceptance/features/webUIActivity/comment.feature
phil-davis: @skshetry please raise an issue to ask about this.
The test can stay like it is, because it demonstrates the current behavior, which might be correct (or not)

Issue #700, opened 2019-02-04
Cannot build


PVince81:
### Steps

1. `git checkout release-2.4.2`
1. `make dist`

### Expected

Successful build

### Actual

Error:
```
± % make dist
rm -Rf /home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity/build/dist
rm -Rf /home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity/build/dist/activity; mkdir -p /home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity/build/dist/activity
cp -R appinfo css img js l10n lib templates README.md CHANGELOG.md CONTRIBUTING.md /home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity/build/dist/activity
/home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity/../../occ integrity:sign-app --privateKey="/home/vincent/.owncloud/certificates/activity.key" --certificate="/home/vincent/.owncloud/certificates/activity.crt" --path="/home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity/build/dist/activity"
ownCloud is not installed - only a limited number of commands are available
Successfully signed "/home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity/build/dist/activity"
tar -czf /home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity/build/dist/activity.tar.gz -C /home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity/build/dist activity
tar -cjf /home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity/build/dist/activity.tar.bz2 -C /home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity/build/dist activity
make appstore
make[1]: Entering directory '/home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity'
rm -rf 
mkdir -p 
mkdir: missing operand
Try 'mkdir --help' for more information.
make[1]: *** [Makefile:121: appstore] Error 1
make[1]: Leaving directory '/home/vincent/Private/Work/workspace/owncloud-for-signing/apps/activity'
make: *** [Makefile:115: dist] Error 2
```

@phil-davis @paurakhsharma please have a look

PVince81:
note: this is blocking the release of the activity app. the release itself isn't urgent but would be good to solve this problem this week

phil-davis:
[...]

PR #708, opened 2019-02-12
Remove duplicate make dist target

phil-davis:
Issue #700 

Somehow this code was added in #660 - see comment https://github.com/owncloud/activity/pull/660/files#r255856913
It duplicated the existing ``make dist`` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,28 +97,6 @@ endif
 # dist
 #
 
-# Builds the source and appstore package
-.PHONY: dist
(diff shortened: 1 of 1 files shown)


phil-davis:
@paurakhsharma do you have any memory about this?
@DeepDiver1975 I see that you were involved in some review of #660 and that various other code blocks were added to ``Makefile`` - do you have any memory about it?

PR #711, opened 2019-02-25
use p() instead of echo

butonic:
we are only printing a description, no need to be able to output raw html here.


karakayasemi:
Seems like this PR introduced regression: 
image

PVince81:
fantastic... revert ?

karakayasemi:
> fantastic... revert ?

Reverted in here: #725

PR #712, opened 2019-03-01
[poc test] Use common makefiles

phil-davis:


phil-davis:
See #714 for later example.

PR #717, opened 2019-03-13
Drop PHP 5.6

phil-davis:
core stable10 no longer supports PHP 5.6
https://github.com/owncloud/core/pull/34698

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -211,18 +211,10 @@ services:
 matrix:
   include:
     # owncloud-coding-standard
-    - PHP_VERSION: 5.6
+    - PHP_VERSION: 7.0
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -211,18 +211,10 @@ services:
 matrix:
   include:
     # owncloud-coding-standard
-    - PHP_VERSION: 5.6
+    - PHP_VERSION: 7.0
       TEST_SUITE: owncloud-coding-standard
 
     # Unit Tests
-    - PHP_VERSION: 5.6
```
phil-davis: This matrix entry was already repeated for PHP 7.0 below.

PR #718, opened 2019-03-13
Remove php 5.6

phil-davis:
In discussion, it has been pointed out that there are more places to remove mention of PHP 5.6:

1) root `composer.json` of the app - bump `platform` `php` to `7.0.8` and then do `composer update` to get an updated `composer.lock` file.

2) `vendor-bin/behat/composer.json` - bump `platform` `php` to `7.0.8` - `composer.lock` is not configured, so no need to touch that, it is auto-generated when tests start. This will only be in apps that have Behat acceptance tests

3) bump the ownCloud server `min-version` in `appinfo/info.xml` to `10.2` - because actually we are running automated testing against the core `stable10` QA tarball, which will next become `10.2`. So we are not automatically sure that the app still works with core releases `10.0.*` or `10.1.*` (e.g. maybe some app code gets added/changed and the developer uses PHP 7.0-only syntax that does not work on PHP 5.6)

Points (1) and (2) seem straightforward and "have to happen"

Point (3) could be avoided if we have a solution.process for https://github.com/owncloud/QA/issues/611 
which ensures that apps are tested against any older supported server versions prior to app release (and maybe every night in the automated testing)


composer.lock
```diff
@@ -45,1432 +45,6 @@
                 "MIT"
             ],
             "time": "2017-09-11T13:13:58+00:00"
-        },
-        {
-            "name": "doctrine/instantiator",
```
PVince81: some stuff just vanished ?

phil-davis: `composer update` decides what it thinks is still needed or not.
I suspect that a lot of these apps have not had anyone do `composer update` for a while - so `composer` will find newer versions of stuff, and dependencies down in the tree that are no longer relevant.

PVince81: can you confirm your statement by doing a `composer update` before your change and see if the same happens ?

phil-davis:
```
git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
phil@bijay-OptiPlex-3050:~/git/owncloud/core/apps-external/activity$ git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
phil@bijay-OptiPlex-3050:~/git/owncloud/core/apps-external/activity$ composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Nothing to install or update
Writing lock file
Generating autoload files
phil@bijay-OptiPlex-3050:~/git/owncloud/core/apps-external/activity$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   composer.lock

no changes added to commit (use "git add" and/or "git commit -a")
phil@bijay-OptiPlex-3050:~/git/owncloud/core/apps-external/activity$ git diff
diff --git a/composer.lock b/composer.lock
index 9c1840d..99b6ac9 100644
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "ef472d2c739b6bf2ebf50aa4af37bc3c",
+    "content-hash": "1a26761260b076ff83b8d6a93fd05b24",
     "packages": [],
     "packages-dev": [
         {
@@ -45,1432 +45,6 @@
                 "MIT"
             ],
             "time": "2017-09-11T13:13:58+00:00"
-        },
-        {
-            "name": "doctrine/instantiator",
-            "version": "1.0.5",
-            "source": {
-                "type": "git",
-                "url": "https://github.com/doctrine/instantiator.git",
-                "reference": "8e884e78f9f0eb1329e445619e04456e64d8051d"
-            },
-            "dist": {
-                "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/instantiator/zipball/8e884e78f9f0eb1329e445619e04456e64d8051d",
-                "reference": "8e884e78f9f0eb1329e445619e04456e64d8051d",
-                "shasum": ""
-            },
-            "require": {
-                "php": ">=5.3,<8.0-DEV"
-            },
-            "require-dev": {
-                "athletic/athletic": "~0.1.8",
-                "ext-pdo": "*",
-                "ext-phar": "*",
-                "phpunit/phpunit": "~4.0",
-                "squizlabs/php_codesniffer": "~2.0"
-            },
...
```
Yes, lots of stuff is removed from `composer.lock` just by doing `composer update` in the `master` branch.

PR #720, opened 2019-03-22
PHPunit 5 to 6 preparation

phil-davis:
https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-6.0.0

1) Adjust `PHPUnit_Framework_*` to be `PHPUnit\Framework\*` (this works now in PHPUnit5 also)
"PHPUnit's units of code are now namespaced. For instance, PHPUnit_Framework_TestCase is now PHPUnit\Framework\TestCase"

2) Add `assertNull()` wrapper to tests that call a method that is expected to return nothing.
PHPUnit6 does more reporting of "risky" tests. A test that has no assertion is considered "risky". Sometimes there are valid tests like that - a test that calls some method, the method has no return and the only external behavior is that the method will throw an exception if it "does not work". In that case, the method "returning" `null` is the indication that it "worked.
This will avoid these tests being reported as "risky".

This passes on PHPUnit5, so could be merged [prior to|in preparation for] updating to PHPUnit6.

### DIFF
--- a/tests/Unit/ConsumerTest.php
+++ b/tests/Unit/ConsumerTest.php
@@ -34,10 +34,10 @@ class ConsumerTest extends TestCase {
 	/** @var \OCA\Activity\Consumer */
 	protected $consumer;
 
-	/** @var \OCA\Activity\Data|\PHPUnit_Framework_MockObject_MockObject */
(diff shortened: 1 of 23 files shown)


phil-davis:
This is an example of the minor stuff that could be done to each app ready for PHPUnit6

PR #725, opened 2019-04-17
Revert "use p() instead of echo"

### LINKED ISSUE #724: Personal activity settings page is not properly rendering html elements
### Steps to reproduce
Open the general personal panel and see Activity section. When I revert https://github.com/owncloud/activity/pull/711, it is working as expected.

### Expected behaviour
Tell us what should happen
image

### Actual behaviour
image

karakayasemi:
#711 introduced regression on html rendering. This PR  reverts #711. 
Fixes #724

### DIFF
--- a/templates/personal.php
+++ b/templates/personal.php
@@ -37,7 +37,7 @@ style('activity', 'settings');
 				</td>
 				<?php endforeach; ?>
 				<td class="activity_select_group" data-select-group="<?php p($activity) ?>">
-					<?php p($data['desc']); ?>
+					<?php print_unescaped($data['desc']); ?>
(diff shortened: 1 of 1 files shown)


PVince81:
mind trying with `print_unescaped()` which is another function of our like `p()`, not sure if that function does anything else than `echo` though

karakayasemi:
Changed it with `print_unescaped`. Also, to learn what it does differently than `echo`, I quickly checked `print_unescaped` function. It is only calling `print` function and as far as I read `echo` and `print` more or less the same. The only significant difference is `print` has a return value.

PR #744, opened 2019-06-24
URL Formatter

kulmann:
The idea of this issue https://github.com/owncloud/customer_portal/issues/411 was, to be able to notify Case owners in customer portal, when users uploaded files to case data folder. The notification needs context, so we wanted to add links to the Case details page and the Account details page. This is not possible with the `BaseFormatter`, as we can't inject a link into the href attribute of a prebuilt a-tag (url gets surrounded with a `strong`-tag). Injecting the complete a-tag is not possible as well, because the BaseFormatter sanitizes HTML.

This PR adds a URL Formatter, which expects an associative array, json encoded, to fulfill the parameter type `string` of `$parameters`, and converts it into an `a`-tag. Keys of the array can be `url`, `name` and `target`. `url` is required, `name` is recommended (so that the links don't look bad). Ideas for improvements are very welcome.

Our notification now looks like this (with fake data of course):
Bildschirmfoto 2019-06-23 um 11 17 55

### DIFF
--- a/lib/Formatter/UrlFormatter.php
+++ b/lib/Formatter/UrlFormatter.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * @author Benedikt Kulmann <b@kulmann.biz>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Activity\Formatter;
+
+use OCP\Activity\IEvent;
+use OCP\Util;
+
+class UrlFormatter implements IFormatter {
+
+	/**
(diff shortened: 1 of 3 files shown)


lib/Formatter/UrlFormatter.php
```diff
@@ -0,0 +1,69 @@
+<?php
+/**
+ * @author Benedikt Kulmann <b@kulmann.biz>
+ *
+ * @copyright Copyright (c) 2019, ownCloud, Inc.
```
DeepDiver1975: GmbH please

tests/Unit/Formatter/UrlFormatterTest.php
```diff
@@ -0,0 +1,95 @@
+<?php
+/**
+ * @author Benedikt Kulmann <b@kulmann.biz>
+ *
+ * @copyright Copyright (c) 2019, ownCloud, Inc.
```
DeepDiver1975: GmbH

tests/Unit/Formatter/UrlFormatterTest.php
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * @param array $methods
+	 * @return IFormatter|\PHPUnit\Framework\MockObject\MockObject
+	 */
+	public function getFormatter(array $methods = []) {
+		if (empty($methods)) {
+			return new UrlFormatter();
+		} else {
+			return $this->getMockBuilder('OCA\Activity\Formatter\UrlFormatter')
```
DeepDiver1975: UrlFormatter::class

tests/Unit/Formatter/UrlFormatterTest.php
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * @dataProvider dataFormat
+	 *
+	 * @param string $parameter
+	 * @param string $expected
+	 */
+	public function testFormat($parameter, $expected) {
+		/** @var \OCP\Activity\IEvent|\PHPUnit\Framework\MockObject\MockObject $event */
+		$event = $this->getMockBuilder('OCP\Activity\IEvent')
```
DeepDiver1975: IEvent::class

Issue #752, opened 2019-07-27
Acceptance test restore.feature:269 failing


phil-davis:
e.g. https://drone.owncloud.com/owncloud/activity/555/133
```
    And user "user0" has restored the file with original path "textfile0.txt"                                                           # TrashbinContext::elementInTrashIsRestored()
      Exception: TrashbinContext::restoreElement cannot restore from trashbin because no element was found for user user0 at original path textfile0.txt in /var/www/owncloud/server/tests/acceptance/features/bootstrap/TrashbinContext.php:244
      Stack trace:
      #0 /var/www/owncloud/server/tests/acceptance/features/bootstrap/TrashbinContext.php(260): TrashbinContext->restoreElement('user0', 'textfile0.txt')
      #1 [internal function]: TrashbinContext->elementInTrashIsRestored('user0', 'textfile0.txt')
      #2 /var/www/owncloud/server/apps/activity/vendor-bin/behat/vendor/behat/behat/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php(109): call_user_func_array(Array, Array)

--- Failed scenarios:

    /var/www/owncloud/server/apps/activity/tests/acceptance/features/webUIActivityDeleteRestore/restore.feature:269

49 scenarios (48 passed, 1 failed)
525 steps (518 passed, 1 failed, 6 skipped)
```

This started being reported after core PR https://github.com/owncloud/core/pull/35906 improved some checks in `restoreElement`  to make sure that restore-from-trashbin steps were actually at least trying to restore something.

phil-davis:
Also see https://drone.owncloud.com/owncloud/activity/555/267
```
--- FeatureContext has missing steps. Define them with these snippets:

    /**
     * @When the public downloads the last created file using the webUI
     */
    public function thePublicDownloadsTheLastCreatedFileUsingTheWebui()
    {
        throw new PendingException();
    }

    /**
     * @When the public downloads the last created folder using the webUI
     */
    public function thePublicDownloadsTheLastCreatedFolderUsingTheWebui()
    {
        throw new PendingException();
    }
```
The step text of those needs adjusting to match changes in core yesterday.

phil-davis:
ToDo:
- investigate scenario that ffails on OC 10.2 https://drone.owncloud.com/owncloud/activity/563/280
```
[...]

PR #753, opened 2019-07-27
Adjust test steps because of core 35926

phil-davis:
Issue #752 

- adjust test step text to match what changed in core (1st commit)
- adjust "delete-restore" and "unshare" test scenarios to do more correct stuff and pass. The problem with `restore.feature:269` started to be actually reported as a fail because of extra checks added to core `TrashbinContext` `restoreElement` (2nd and 3rd commit)
- skip the moved scenario on OcV10.2 (we can investigate later, let's get the passing things merged)

### DIFF
--- a/tests/acceptance/features/webUIActivityDeleteRestore/restore.feature
+++ b/tests/acceptance/features/webUIActivityDeleteRestore/restore.feature
@@ -248,7 +248,7 @@ Feature: Restored files/folders activities
     Then the activity number 1 should have message "You restored lorem.txt" in the activity tab
     And the activity number 2 should have message "You deleted lorem.txt" in the activity tab
 
-  Scenario: Sharer and sharee check activity after sharer deletes shared file and then again restore it
+  Scenario: Sharer and sharee check activity after sharer deletes a shared file and then restores it
     Given these users have been created with default attributes and without skeleton files:
       | username |
       | user1    |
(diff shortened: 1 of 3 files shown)


tests/acceptance/features/webUIActivityDeleteRestore/restore.feature
```diff
@@ -266,17 +266,20 @@ Feature: Restored files/folders activities
     And the activity number 2 should contain message "User Zero deleted textfile0.txt" in the activity page
     And the activity number 3 should have a message saying that user "User Zero" has shared "textfile0.txt" with you
 
-  Scenario: Sharer and sharee check activity after sharee deletes shared file and then again restore it
+  Scenario: Sharer and sharee check activity after sharee deletes a shared file and then restores it
```
phil-davis: This scenario is adjusted so that a folder is shared, and then a file in the folder is deleted and restored.
Previously it tried to "delete" a single received shared file. Actually that "delete" is really an "unshare" of the received file. The "restore" that the scenario previously did, did not actually work (because there was no proper "delete" to be restored). The previous scenario had only been passing because it did not test for "restore" activity entries anyway.

tests/acceptance/features/webUIActivitySharingInternal/sharingInternal.feature
```diff
@@ -286,6 +286,21 @@ Feature: Sharing file/folders activities
     And user "user1" has deleted the last share
     And the user browses to the activity page
     Then the activity number 1 should have a message saying that "User One" removed the share of "User One" for "simple-folder"
+
+  Scenario: Sharer and sharee check activity after sharee unshares a shared file
```
phil-davis: This scenario is added here. It is similar to the original scenario that was in `restore.feature`
It tests the case when a sharee receives just a share of a single file, and unshares it.

PR #755, opened 2019-08-08
Trigger CI to test new drone

phil-davis:


phil-davis:
CI passes - that is a good start :)

PR #756, opened 2019-08-09
Start to integrate a reference starlark pipeline

tboerger:


phil-davis:
I added 2 small commits. It looks good now. 

I think there are some webUI test scenarios that are failing on firefox. We have not previously run the activity app webUI tests on firefox, so that is a "new feature". I will look at the results. I can then take one of the following paths:

1) fix the test scenarios so they do pass on firefox, or;
2) skip "a few" test scenarios that fail on firefox and make an issue to look into it (if fixing now seems too hard), or:
3) remove all firefox matrix entries (if the firefox fails are that bad)

phil-davis:
https://drone.owncloud.com/owncloud/activity/605/20/9
```
Scenario: Uploading new file using async should register in the activity list # /var/www/owncloud/server/testrunner/apps/activity/tests/acceptance/features/webUIActivityCreateUpdate/createFilesFolder.feature:135
Given the administrator has enabled async operations # FeatureContext::triggerAsyncUpload()
And using new DAV path # FeatureContext::usingOldOrNewDavPath()
And user "user0" has uploaded the following chunks asynchronously to "/text.txt" with new chunking # FeatureContext::userHasUploadedTheFollowingChunksUsingNewChunking()
| 1 | AAAAA |
| 2 | BBBBB |
| 3 | CCCCC |
HTTP status code is not the expected value
Failed asserting that an array contains 202.
When the user browses to the activity page # WebUIActivityContext::userHasBrowsedToTheActivityPage()
Then the activity number 1 should contain message "You created text.txt" in the activity page # WebUIActivityContext::theActivityNumberShouldContainMessageInTheActivityPage()
SCENARIO RESULT: (fail) 
```
is due to changes made in core to better check that `Given` steps have worked.
Unfortunately the checks were too good! Async chunked uploads have API calls that return `202` "accepted".

That is being fixed in core by PR https://github.com/owncloud/core/pull/36042 and will come in the nightly QA tarball.

tboerger:
Thank you very much for these catches. In worst case I can simply drop the firefox acceptance tests, I just thought it could make sense.

tboerger:
Only one chrome pipeline is still failing: https://drone.owncloud.com/owncloud/activity/611/28/12

phil-davis:
There were some other unrelated core acceptance test changes yesterday. I am working on getting the related changes needed in activity, then I will rebase this...

phil-davis:
IMO this is good now. The commits that I added are:

Change TEST_SERVER to TEST_SERVER_URL - bugfix

Change to selenium/standalone-chrome-debug:3.141.59-oxygen - bugfix

substitute app name in testrunner copy - remove a last remaining hardcoded 'activity' down in the "library" methods, and substitute `config['app']` instead. That completes abstracting out the app name.

use testrunner for all acceptance tests - like PR #760 - today @individual-it and I discussed the inconsistency about the way that we had been getting the core test runner code from github when testing against a release like 10.2.1 but using the test runner code inside the daily-master-qa tarball when testing against that QA tarball. That difference is not actually needed, and sometimes creates problems during days when github core master test code has changed and the QA tarball is stale.

Skip tests that do not scroll into view on Firefox - there are some stupid things with the Firefox webdriver... and it does not automatically scroll an element (checkbox in this case) into view before trying to click it. It will take time to sort out the more "manual" scrolling for the failing tests. So I skipped a few test scenarios for now. But 99% of the webUI scenarios run and pass on Firefox!

move testrunner to own folder - I am not sure why we put the testrunner in a folder under where the server-under-test is installed `/var/www/owncloud/server/testrunner` (probably somewhere back in the mists of time there was a priv problem...) Doing that pollutes the `/var/www/owncloud/server` folder of the server-under-test. Now the testrunner goes in `/var/www/owncloud/testrunner`

.drone.starlark
```diff
@@ -0,0 +1,606 @@
...
+			],
+			'federate': False
+		},
+		'apiActivity': {
+			'server': [
+				'daily-master-qa',
+				'10.2.1'
+			],
+			'browser': [
+				'headless'
```
individual-it: what about having `null` or `none` here. Its confusing to have `headless` because there are headless browsers, but here we do not start any browser

.drone.starlark
```diff
@@ -0,0 +1,606 @@
...
+			pipelines.append(result)
+
+	return pipelines
+
+def acceptance():
+	pipelines = []
+
+	for suite, matrix in config['acceptance'].items():
+		for server in matrix['server']:
+			for browse in matrix['browser']:
```
individual-it: typo?
```suggestion
			for browser in matrix['browser']:
```

.drone.starlark
```diff
@@ -0,0 +1,606 @@
...
+				name = 'unknown'
+				php = '7.0'
+
+				if suite.startswith('webUI'):
+					name = 'webui-%s-%s-%s' % (suite.replace('webUI', ''), browse, server.replace('daily-', '').replace('-qa', ''))
+
+				if suite.startswith('api'):
+					name = 'api-%s-%s-%s' % (suite.replace('api', ''), browse, server.replace('daily-', '').replace('-qa', ''))
```
individual-it: I don't think we need to mention the browser here, as its an API test
```suggestion
					name = 'api-%s-%s' % (suite.replace('api', ''), server.replace('daily-', '').replace('-qa', ''))
```

.drone.starlark
phil-davis: Changed to `none` and removed it from the generated matrix name for the api case.

.drone.starlark
phil-davis: I renamed various functions, including `browser()` to (hopefully) improve understanding. And also so that `browser` is no longer function and can be used as a variable name here.

.drone.starlark
phil-davis: Removed from name.

phil-davis:
This code and later has been implemented in PR #768

PR #760, opened 2019-08-16
Always use testrunner for acceptance tests

phil-davis:
Previously we have been:
1) using the QA tarball acceptance test code when testing the app with the `daily-master-qa` tarball.
2) using the latest core master acceptance test code cloned from git when testing the app with the `10.2.1` or other release tarballs.

This sometimes creates inconsistency when core test steps have been modified, added, deleted in git core master but the `daily-master-qa` tarball is out-of-date. In that case it can be difficult or impossible to make the app CI pass - because some gherkin step text needs to be one way for `daily-master-qa` and another way for git core master.

Actually we may as well just always use the "testrunner" logic no matter which core tarball is the system-under-test.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -28,7 +28,7 @@ pipeline:
       - make vendor-bin-deps
     when:
       matrix:
-        USE_RELEASE_TARBALL: true
+        TEST_SUITE_TYPE: acceptance
 
   install-fed-server:
     image: owncloudci/core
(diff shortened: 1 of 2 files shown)


phil-davis:
Note: this could be later optimized to avoid having to `make` all the core dependencies in the testrunner clone of core. i.e. getting rid of:
```
      - cd /var/www/owncloud/server/testrunner
      - make install-composer-deps
      - make vendor-bin-deps
```
but that is more complicated! I can work on that later.

PR #762, opened 2019-08-16
Separate starlark config

phil-davis:
@tboerger the config array can be put into a separate file like this.
At least that will make the main `.drone.starlark` "code library" be a file that can be the same in [all|most] app repos.

Any better ideas for separating common code from app-specific config data?

Will something like this still work when starlark server integration comes?


tboerger:
I would keep that in a single file as it's unclear if Drone will natively support imports or not.

phil-davis:
This is a PoC. It will be easy to do something along these lines later if we [can|want]

PR #763, opened 2019-08-19
Refactor starlark 1

phil-davis:
1) Provide defaults for acceptance test matrix - to avoid duplicating the common matrix combination pattern of acceptance tests

2) do not install the federation sharing for `webUIActivitySharingInternal` - that suite should not need it.


phil-davis:
closing - drone is not starting

PR #771, opened 2019-08-29
Support useBundledApp, refactor testrunner

phil-davis:
This PR adds support in the drone starlark for the following things that are useful for some apps that also are bundled into core release tarballs:

1) `useBundledApp` - if true, then assume that the app should already be bundled in the core tarball under test. (i.e. do not put the app GitHub branch into the server under test). If false then exclude the app when fetching and unpacking the core tarball, and put the app GitHub branch into the server under test. The default for `useBundledApp` is false.

2) refactor `fixPermissions` so that it just fixes the `www-data` file owner for the actual `server`  and `federated` folders that are  running under Apache. This avoids messing with permissions of the `testrunner` directory.

3) refactor so that the PR under test goes first to the `testrunner` directory `/var/www/owncloud/testrunner/apps/appname` and is only copied into the sserver-under-test in the case when `useBundledApp`  is false.

4) refactor tyhe `appInstallCommand` code so that it does not need the strange `else ':'` syntax (that I used at first to insert a "noop" bash command)

The code here has been tried in `configreport` (which is a bundled app) and it works when `useBundledApp` is either `True` or `False`

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -78,15 +78,17 @@ def main(ctx):
 	return before + stages + after
 
 def beforePipelines():
-	return codestyle() + javascript() + phpstan() + phan()
+	return codestyle() + phpstan() + phan()
 
 def stagePipelines():
-	phpunitPipelines = phpunit()
+	jsPipelines = javascript()
+	phpunitPipelines = phptests('phpunit')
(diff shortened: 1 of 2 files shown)


phil-davis:
It had a drone network error - I restarted the drone CI build.

PR #782, opened 2019-09-13
Update starlark code

phil-davis:
These are additional features found to be needed/useful when implementing drone in `files_antivirus`

1) Refactored the processing of defaults to remove a lot of similar copy-paste statements for each parameter.

2) Add support for:
`extraSetup` - a step that does any needed extra setup after the app is enabled but before the tests are run.
`extraServices` - definitions of any extra services that need to be running during the tests (e.g. `clamav`, `scality` etc)
`extraEnvironment` - any extra environment variables that need to be defined in the step that runs the tests.
`extraCommandsBeforeTestRun` - any special command that needs to be run in the step that runs the tests, just before the `make test-*` command.

and make these all available in javascript, phpunit and acceptance tests.

3) Allow `extraApps` to specify a command to run after cloning the extra app and before enabling it. (e.g. the command might be `make` or `composer install` or...)

I have confirmed that this code works with all the existing apps that have `.drone.starlark` and generates the same `.drone.yml` as it does now.

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -110,8 +110,8 @@ def codestyle():
 
 	if 'defaults' in config:
 		if 'codestyle' in config['defaults']:
-			if 'phpVersions' in config['defaults']['codestyle']:
-				default['phpVersions'] = config['defaults']['codestyle']['phpVersions']
+			for item in config['defaults']['codestyle']:
+				default[item] = config['defaults']['codestyle'][item]
 
 	codestyleConfig = config['codestyle']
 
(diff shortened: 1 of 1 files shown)


.drone.starlark
```diff
@@ -318,30 +321,29 @@ def javascript():
 	default = {
 		'coverage': False,
 		'logLevel': '2',
+		'extraSetup': None,
 		'extraServices': [],
+		'extraEnvironment': {},
+		'extraCommandsBeforeMake': [],
```
individual-it:
```suggestion
		'extraCommandsBeforeTestrun': [],
```

phil-davis: fixed

PR #783, opened 2019-09-15
Update starlark code

phil-davis:
- support code style for JavaScript
- move `installExtraApps` before `setupServerAndApp` because sometimes the app under test requires some other app to be installed first (e.g. `files_ldap_home` needs `user_ldap` to first be installed before it can be enabled successfully)
- support `extraApps` for `phpunit` (as well as for `acceptance`)
- allow the "category" key to optionally be used in the pipeline matrix name by setting `'includeKeyInMatrixName': True` - this allows the same test suite, or phpunit job to be run with some different other context, without causing duplicate pipeline matrix names.

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -461,6 +461,7 @@ def phptests(testType):
 		'extraServices': [],
 		'extraEnvironment': {},
 		'extraCommandsBeforeTestRun': [],
+		'extraApps': [],
 	}
 
 	if 'defaults' in config:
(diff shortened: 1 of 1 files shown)


phil-davis:
I will just do this PR to the `activity` app. Then I will have current code in this app, which I can use for doing starlark for new app repos.

Some time in the future when I get more... code enhancements, then I will apply them to all the other repos with starlark. It is getting tedious to do it every day!

PR #787, opened 2019-09-27
Drop drone yml

phil-davis:
1) Update the starlark to the latest
2) drop `.drone.yml` - the drone server should read  `.drone.star` now


phil-davis:
@tboerger drone has not triggered off `.drone.starlark`
Is there some magic to be done in the repo settings so it knows to trigger off the `starlark` and not the `yml`?

phil-davis:
I will make a new PR - this is out-of-date

PR #792, opened 2019-10-07
Update drone starlark with latest code 20191011

phil-davis:
Allow for running all acceptance tests in a matrix of `numberOfParts`

This is used by `files_primary_s3` and `encryption` to run core tests.

And other changes from recent days to support LDAP...

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -552,11 +552,13 @@ def phptests(testType):
 		'coverage': True,
 		'includeKeyInMatrixName': False,
 		'logLevel': '2',
+		'cephS3': False,
+		'scalityS3': False,
 		'extraSetup': [],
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -1063,19 +1137,8 @@ def installTestrunner(phpVersion, useBundledApp):
...
-		if extraApps == []:
-			return []
-		apps = {}
-		for app in extraApps:
-			apps[app] = ''
-	else:
-		apps = {}
-		for app, command in extraApps.items():
-			apps[app] = command
-
```
phil-davis: The code is simpler now - `extraApps` can be always a `dict` and so there is no need for special code if it is a `list`

PR #804, opened 2019-11-21
Add flexibility in drone starlark to override default ceph and scality setups

phil-davis:
These enhancements were needed for `objectstore` - that app does its own thing, rather than relying on `files_primary_s3`. So we need to be able to override the default setups that happen when `cephS3` or `scalityS3` is set to true.

So you can do stuff like:
- specify `setupCommands` for ceph or scality to override what is done by default.
- specify `filesPrimaryS3Needed` false to stop the default install/enable of `files_primary_s3`

```
			'cephS3': {
			    'setupCommands': [
				    'wait-for-it -t 60 ceph:80',
				    'cd /var/www/owncloud/server',
				    'cp apps/objectstore/tests/env/s3-ceph.config.php config/autotest-objectstore-s3-ceph.config.php'
			    ],
			    'filesPrimaryS3Needed': False,
			    'createFirstBucket': False
			},
```

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -584,7 +584,23 @@ def phptests(testType):
 		for item in default:
 			params[item] = matrix[item] if item in matrix else default[item]
 
-		if ((config['app'] != 'files_primary_s3') and ((params['cephS3'] != False) or (params['scalityS3'] != False))):
+		cephS3Params = params['cephS3']
+		if type(cephS3Params) == "bool":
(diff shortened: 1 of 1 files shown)


phil-davis:
I have tested putting this code in the apps that use ceph or scality, and it does not change/break the drone config.

`activity` is the first app alphabetically, so I keep the drone starlark code up-to-date here, where I can always find the "latest".

PR #811, opened 2019-12-09
occ command to send all pending notifications

VicDeo:
## Concept
- Create an occ command to send out all pending notifications instantly
- Don't touch existing code

## Implementation:
1,2,3 is repeated in batches of 250 until the count of notified users becomes 0.  

1. we get all uids and all max(mail_id)  per uid. (The queue could be changed when the sending in process so we have max mail_id to work with)
2. we iterate through uids sending an email containing all user notifications for actions with `mail_id <= max(mail_id)`
3. we delete all notifications in a one go as 
```
DELETE FROM *PREFIX*activity_mq WHERE
 (uid='uid1' AND mail_id<=max_mail_id_uid1)
 OR (uid='uid2' AND mail_id<=max_mail_id_uid2)
...
 OR (uid='uidN' AND mail_id<=max_mail_id_uidN)

### DIFF
--- a/lib/Command/SendEmails.php
+++ b/lib/Command/SendEmails.php
@@ -1,0 +1,52 @@
+<?php
+/**
+ * @author Viktar Dubiniuk <dubiniuk@owncloud.com>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Activity\Command;
+
+use OCA\Activity\BackgroundJob\EmailNotification;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: lib/Command/SendEmails.php [hunk 1]
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	/**
+	 * @param InputInterface $input
+	 * @param OutputInterface $output
+	 *
+	 * @return int|void
+	 */
+	public function execute(InputInterface $input, OutputInterface $output) {
+		do {
```
DeepDiver1975: what about user feedback? progress bar or something.
in case this also supposed to be use in cron -> add an option to enable the user feedback .... or use the verbosity level

lib/Command/SendEmails.php
```diff
@@ -0,0 +1,127 @@
...
+			try {
+				$language = (!empty($userLanguages[$uid])) ? $userLanguages[$uid] : $defaultLang;
+				$timezone = (!empty($userTimezones[$uid])) ? $userTimezones[$uid] : $defaultTimeZone;
+				$this->mqHandler->sendAllEmailsToUser($uid, $user['email'], $language, $timezone, $user['max_mail_id']);
+				$sentMailForUsers[] = [
+					'uid' => $uid,
+					'maxMailId' => $user['max_mail_id']
+				];
+			} catch (\Exception $e) {
+				// Delete all entries we dealt with
```
DeepDiver1975: log statement is missing

lib/Command/SendEmails.php
VicDeo: Progress bar is tricky. We work with batches of 250 users so totals may increase while a singe batch is processed. Let me try...

VicDeo: added a simple progress bar without totals

lib/Command/SendEmails.php
VicDeo: added

VicDeo:
@DeepDiver1975 rereview please

PR #814, opened 2019-12-12
Increase wait-for-db timeout from 120 to 600 seconds

phil-davis:
## Description
The drone agents are sometimes taking "a long time" (tm) to pull docker images. In particular this is a problem when it takes a long time to pull the database image (mariadb, postgresql, oracle...). The `install-core` step waits 120 seconds for the database to become available. If the pull took more than 2 minutes, then we get a timeout in `install-core` and the pipeline fails. That means the whole build fails and it has to be started all over again. That is very annoying.

This is seen mostly first thing in the morning, when the autoscaler starts new drone agents. Specially if a lot of drone agents get started at a similar time, they seem to be slow to pull each docker image the first time. After that, they have the docker images cached, so later pulls of the docker images do not have this time delay.

Since we know there is a very variable delay here, set the timeout to a high value (10 minutes = 600 seconds). That will avoid annoying unnecessary fails, while still providing a reasonable time-to-failure in the rare case where the database really did fail on startup.

## How Has This Been Tested?
CI

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1285,7 +1285,7 @@ def setupCeph(serviceParams):
 
 	createFirstBucket = serviceParams['createFirstBucket'] if 'createFirstBucket' in serviceParams else True
 	setupCommands = serviceParams['setupCommands'] if 'setupCommands' in serviceParams else [
-		'wait-for-it -t 120 ceph:80',
+		'wait-for-it -t 600 ceph:80',
 		'cd /var/www/owncloud/server/apps/files_primary_s3',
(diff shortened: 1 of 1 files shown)


phil-davis:
Similar to https://github.com/owncloud/core/pull/36579

Doing this here, so that activity app has the "current" starlark. That way this will get applied to other apps when we next have a reason to update every app.

I will also apply this to a couple of the apps with big/long pipelines, to reduce the impact of this problem.

PR #826, opened 2020-01-10
Drone: Update drone.star to use mysql on federated server as database

dpakach:
Use mysql as database for federated server while running acceptance tests in drone CI.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -848,6 +848,7 @@ def acceptance():
 										environment['S3_TYPE'] = 'ceph'
 									if (params['scalityS3'] != False):
 										environment['S3_TYPE'] = 'scality'
+								federationDbSuffix = '-federated'
 
 								result = {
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -1357,3 +1336,68 @@ def dependsOn(earlierStages, nextStages):
...
+		'name': dbName + suffix,
+		'image': db,
+		'pull': 'always',
+		'environment': {
+			'MYSQL_USER': getDbUsername(db),
+			'MYSQL_PASSWORD': getDbPassword(db),
+			'MYSQL_DATABASE': getDbDatabase(db) + suffix,
+			'MYSQL_ROOT_PASSWORD': getDbRootPassword()
+		}
+	}]
```
phil-davis: add an end-of-line here. That will make it exactly match what is in files_primary_s3

PR #838, opened 2020-02-28
Correct the Oracle Db auth in .drone.star

phil-davis:
Like done in core PR https://github.com/owncloud/core/pull/37031


phil-davis:
Note: this change is not really important to have straight away in all app repos.
`activity` is the first app alphabetically, so we keep the "latest standard" `.drone.star` code here.
When we have other changes to `.drone.star` code, then we can bother applying it to other repos.

PR #842, opened 2020-03-09
Skip disabled users and do not break the cronjob on the first unsent email

VicDeo:
- [x] we don't need to fix DI since https://github.com/owncloud/core/pull/21281
- [x] Do not try to send notification if the user is disabled
- [x] Keep on sending emails after the exception

### DIFF
--- a/lib/BackgroundJob/EmailNotification.php
+++ b/lib/BackgroundJob/EmailNotification.php
@@ -68,25 +68,11 @@ class EmailNotification extends TimedJob {
 		// Run all 15 Minutes
 		$this->setInterval(15 * 60);
 
-		if ($mailQueueHandler === null || $config === null || $logger === null || $isCLI === null) {
-			$this->fixDIForJobs();
-		} else {
-			$this->mqHandler = $mailQueueHandler;
-			$this->userManager = $userManager;
-			$this->config = $config;
-			$this->logger = $logger;
-			$this->isCLI = $isCLI;
-		}
-	}
-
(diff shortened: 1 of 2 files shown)


lib/BackgroundJob/EmailNotification.php
```diff
@@ -144,10 +142,14 @@ protected function runStep($limit, $sendTime) {
 				$this->mqHandler->sendEmailToUser($uid, $user['email'], $language, $timezone, $sendTime);
 				$sentMailForUsers[] = $uid;
 			} catch (\Exception $e) {
-				// Run cleanup before we die
-				$this->mqHandler->deleteSentItems($sentMailForUsers, $sendTime);
-				// Throw the exception again - which gets logged by core and the job is handled appropriately
```
micbar: Why is the cleanup not needed anymore?

micbar:
One question. Rest looking fine

lib/BackgroundJob/EmailNotification.php
VicDeo: we don't rethrow exception any more which means that we always reach the line 155 now
https://github.com/owncloud/activity/pull/842/files#diff-35fc1857b1cf16ed158a97c77089f80cL155

micbar: ok, didn't see that,
👍

micbar:
Coverage is miscalculating

@VicDeo Can I merge now and override?

PR #850, opened 2020-04-21
[Tests-Only] Run acceptance tests with php7.4

dpakach:
Running acceptance tests with php7.4 in CI.

Part of https://github.com/owncloud/QA/issues/645


phil-davis:
LGTM - drone has used PHP 7.4 for the server-under-test and the test-runner.
https://drone.owncloud.com/owncloud/activity/1220/11/10

PR #875, opened 2020-07-17
if object_name empty for file activity, derive file path from filecache info path

mrow4a:
For some activities, no object_name is stored (for rename, tags, comment, tag). Thus path cannot be derived from object_name and preview are wrongly shown:

Zrzut ekranu 2020-07-17 o 17 31 28

This PR fixes it

Zrzut ekranu 2020-07-26 o 11 08 50

- [x] fix issue
- [x] add tests

### DIFF
--- a/lib/Controller/OCSEndPoint.php
+++ b/lib/Controller/OCSEndPoint.php
@@ -317,7 +317,7 @@ class OCSEndPoint {
 					], '', '&');
 
 					$preview['source'] = $this->urlGenerator->linkTo('', 'remote.php')
-						. '/dav/files/' . \rawurlencode($this->user) . \OCP\Util::encodePath($filePath)
+						. '/dav/files/' . \rawurlencode($this->user) . \OCP\Util::encodePath($info['path'])
 						. "?$query";
 				} else {
 					$preview['source'] = $this->urlGenerator->linkToRoute('core_ajax_preview', [
--- a/tests/Unit/Controller/OCSEndPointTest.php
+++ b/tests/Unit/Controller/OCSEndPointTest.php
@@ -563,10 +563,17 @@ class OCSEndPointTest extends TestCase {
 
 	public function dataGetPreview() {
 		return [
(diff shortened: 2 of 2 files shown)


lib/Controller/OCSEndPoint.php
```diff
@@ -316,6 +316,11 @@ protected function getPreview($owner, $fileId, $filePath) {
...
 					], '', '&');
 
+					if ($filePath === '') {
+						// if path was not provided by activity (object_name empty for given object_id)
+						// derive path from file info
+						$filePath = $info['path'];
+					}
 					$preview['source'] = $this->urlGenerator->linkTo('', 'remote.php')
 						. '/dav/files/' . \rawurlencode($this->user) . \OCP\Util::encodePath($filePath)
 						. "?$query";
```
mrow4a, line 326: @DeepDiver1975 I tried also just using `$info['path']` instead of `filePath`, but commented test fails

tests/Unit/Controller/OCSEndPointTest.php
```diff
@@ -566,6 +566,7 @@ public function dataGetPreview() {
 			['author', 42, '/path', '/currentPath', true, true, false, '/preview/dir', true],
 			['author', 42, '/file.txt', '/currentFile.txt', false, true, false, '/preview/mpeg', true],
 			['author', 42, '/file.txt', '/currentFile.txt', false, true, true, 'remote.php/dav/files//file.txt?preview=1&x=150&y=150', false],
```
mrow4a, line 568: @DeepDiver1975 how is this possible that activity is saved with `object_name` being `/file.txt`, but fileinfo returns `/currentFile.txt` ? This is e.g. in case of rename ? Why in this test we assert for `object_name` (`$path = /file.txt`) though ?

mrow4a:
@DeepDiver1975 could you re-review ?

karakayasemi:
Rebased to be able to merge

PR #893, opened 2020-10-14
 files/folders in html emails are now clickable

DeepDiver1975:
Fixes https://jira.owncloud.com/browse/OC-2285

## Screenshot

Screenshot_2020-10-14 MailHog

### Url Format
`<a href="http://localhost/index.php/f/127">Screenshot_2020-02-26 Document Classification app ownCloud marketplace.png</a>`

### DIFF
--- a/lib/HtmlTextParser.php
+++ b/lib/HtmlTextParser.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * @author Thomas Müller <1005065+DeepDiver1975@users.noreply.github.com>
+ *
+ * @copyright Copyright (c) 2020, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


micbar:
@DeepDiver1975 Thanks.

I see that you create the links to the owncloud page.
In the concept we thought about using the private link.

DeepDiver1975:
> I see that you create the links to the owncloud page.
> In the concept we thought about using the private link.

The problem is that I don't have the private link at hand at this point in time ..... 
..... I can do some more refactoring if this is a must have - which might cause some :boom:

micbar:
@DeepDiver1975 So far so good.
Tested it, works.

DeepDiver1975:
> ..... I can do some more refactoring if this is a must have - which might cause some boom

no refactoring was necessary - just some :unicorn: dust ...

micbar:
Regex unicorn power

micbar (approved):
LGTM

PR #904, opened 2020-11-30
[Tests-Only] improve starlark code for acceptance tests

dpakach:
Improve starlark code for acceptance tests.


phil-davis (approved):
That gets rid of the huge indent - good.

PR #908, opened 2020-12-07
[Tests-Only][CI] Set env vars to empty strings instead of None

dpakach:
Update starlark to use empty strings instead of None
This was causing error in `drone starlark generate` which we do in the user_ldap app


phil-davis (approved):

phil-davis:
@dpakach I also noticed `sonar.pullrequest.github.repository` in `sonar-project.properties` is still set to `owncloud/core`
I think this might be why we are not getting SonarCloud analysis posting back to the PR.
I fixed it in this PR - let's see what happens to CI.

PR #916, opened 2020-12-30
[docs-only] SonarCloud badges

phil-davis:
This has the Quality Gate Status, Security Rating and Codecov badges.

@micbar are these the "standard" badges that we want to put?

When we have decided on the "standard" then I can get that done in all repos.


micbar:
I would add `security rating`

phil-davis:
> I would add `security rating`

done

micbar (approved):

PR #910, opened 2020-12-08
Bump karma-jasmine from 3.1.1 to 4.0.1

phil-davis:
Bumps [karma-jasmine](https://github.com/karma-runner/karma-jasmine) from 3.1.1 to 4.0.1.
- [Release notes](https://github.com/karma-runner/karma-jasmine/releases)
- [Changelog](https://github.com/karma-runner/karma-jasmine/blob/master/CHANGELOG.md)
- [Commits](https://github.com/karma-runner/karma-jasmine/compare/v3.1.1...v4.0.1)

Signed-off-by: dependabot-preview[bot] <support@dependabot.com>

This is a cherry-pick from PR #882 - drone CI is not starting there.

PR #917, opened 2021-01-04
[tests-only] Use nodejs docker for JS tests

phil-davis:
1) Use `owncloudci/nodejs:12` for JS  tests in CI to avoid stuff like https://drone.owncloud.com/owncloud/activity/1662/2/4
```
error karma@5.2.3: The engine "node" is incompatible with this module. Expected version ">= 10". Got "8.17.0"
```

Why have we been using `owncloudci/php` for this anyway? I don't know. It has https://github.com/owncloud-ci/php/blob/master/v7.2/Dockerfile.amd64 - `nodejs` v8 comes "by default" with Ubuntu 18.04, and that `nodejs` is getting too old.

2) Remove the check for `composer` from `Makefile`. `make test-js` does not need composer, and probably other `make` targets do not need it either.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -586,7 +586,7 @@ def javascript(ctx):
 		[
 			{
 				'name': 'js-tests',
-				'image': 'owncloudci/php:7.2',
+				'image': 'owncloudci/nodejs:12',
 				'pull': 'always',
 				'environment': params['extraEnvironment'],
 				'commands': params['extraCommandsBeforeTestRun'] + [
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,7 @@
 #
-# Define NPM and COMPOSER_BIN and check if they are available on the system.
+# Define NPM and check if it is available on the system.
 #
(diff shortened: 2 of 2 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/activity/1664/2/4
```
04 01 2021 09:01:08.014:ERROR [phantomjs.launcher]: /var/www/owncloud/server/apps/activity/node_modules/phantomjs-prebuilt/lib/phantom/bin/phantomjs: error while loading shared libraries: libfontconfig.so.1: cannot open shared object file: No such file or directory
```

It looks like `owncloudci/nodejs` is missing other stuff that is handy for running JS tests. Needs sorting out - ether get a later `nodejs` in `owncloudci/php` or add more stuff to `owncloudci/nodejs` or "manually" install more stuff in drone CI...

phil-davis:
Using the latest PHP gets a later Ubuntu which has a later `nodejs`, so I do that in PR #910

PR #942, opened 2021-03-31
Automations in activity stream

JanAckermann:
These adjustments properly handle the activity stream for actions that were triggered by an automation (like the workflow app e.g.). They require the changes from https://github.com/owncloud/core/pull/38605 and https://github.com/owncloud/workflow/pull/255 to work.

Needed for: https://github.com/owncloud/enterprise/issues/4222

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -45,6 +45,9 @@ config = {
 				'chrome',
 				'firefox'
 			],
+			'servers': [
+				'daily-master-qa',
+			],
 		},
 		'federatedSuites': {
 			'suites': {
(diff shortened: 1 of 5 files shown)


JanAckermann (commented):
Lorem ipsum

.drone.star
```diff
@@ -45,6 +45,9 @@ config = {
 				'chrome',
 				'firefox'
 			],
+			'servers': [
```
JanAckermann, line 48: **Restore after release!**

.drone.star
```diff
@@ -60,6 +63,9 @@ config = {
 			'suites': [
 				'apiActivity'
 			],
+			'servers': [
```
JanAckermann, line 66: **Restore after release!**

jvillafanez (approved):

phil-davis (approved):
drone CI changes are OK and necessary.
I will think about where we remember these, so that we put the test matrix back after release of core 10.7.1

phil-davis:
After release of core, we can search all apps for `min-version="10.7.1"` - then find the changes to `.drone.star` and reverse them.

PR #944, opened 2021-04-12
Add shareExpire-param to indicate if a share was removed due to its expiry

JammingBen:
Goes together with https://github.com/owncloud/core/pull/38631 (although the changes do not heavily depend on each other, so no braking change).

Fixes https://github.com/owncloud/enterprise/issues/4455


jvillafanez:
I'd check if `$params['shareExpired']` is present in order to keep backwards compatibility, otherwise we'll have to increase the minimum core version for the app.

jvillafanez (approved):

PR #946, opened 2021-05-06
[tests-only] Add ability to skip pipelines in CI

saw-jan:
## Description
This PR adds the ability to skip certain CI pipelines (`PhpUnitTests`, `JavascriptTest` and `acceptanceTest`)

## Related Issue
- part of  https://github.com/owncloud/QA/issues/670

PR #947, opened 2021-05-06
[tests-only] skip some tests

phil-davis:
To check that the `skip` settings in `.drone.star` are working. Test of PR #946


phil-davis:
I have done a few combinations of skipping things. They all skipped what I expected, and ran the rest of the pipelines.
:rocket:

PR #949, opened 2021-05-11
[tests-only] Feature to debug specific test suites in CI

saw-jan:
This PR adds a feature that enables us to run specific test suites in CI

part of https://github.com/owncloud/QA/issues/670

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -867,7 +867,9 @@ def acceptance(ctx):
 		'numberOfParts': 1,
 		'cron': '',
 		'pullRequestAndCron': 'nightly',
-		'skip': False
+		'skip': False,
+		'debugSuites': [],
+		'skipExceptParts': []
 	}
 
 	if 'defaults' in config:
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -48,6 +50,8 @@ config = {
 			'servers': [
 				'daily-master-qa',
 			],
+			'skip': False,
+			'debugSuites': []
```
kiranparajuli589, line 54: as these are also in defaults, why they are needed to specify here?

.drone.star
```diff
@@ -882,6 +884,14 @@ def acceptance(ctx):
 				suites[suite] = suite
 		else:
 			suites = matrix['suites']
+		
+		if 'debugSuites' in matrix and len(matrix['debugSuites']) != 0:
+			if type(matrix['suites']) == "list":
```
phil-davis, line 889:
```suggestion
			if type(matrix['debugSuites']) == "list":
```

we do not want to assume that `suites` and `debugSuites` are always the same type.

phil-davis (approved):

PR #951, opened 2021-05-30
[tests-only] Adjust test step for deleting files

phil-davis:
https://drone.owncloud.com/owncloud/activity/1899

The step was removed in core, because there was a similar step already existing. Use that API step to delete files.


individual-it (changes requested):
header of the table needs to be changed
`Row with header 'path' expected to be in table but not found (Exception)`

phil-davis:
@individual-it fixed

individual-it (approved):

PR #958, opened 2021-06-10
Update coding-standard to major version 3

phil-davis:
Part of https://github.com/owncloud/coding-standard/issues/22

The similar style change in core is PR https://github.com/owncloud/core/pull/38828

This can be done to each app "some time" (no hurry). It does not break or change any behavior, so it can happen and the new code can be released whenever there happens to be an app release - it is not dependent on anything.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,8 @@ vendor/
 tests/js/package-lock.json
 
 # php-cs-fixer
+.php-cs-fixer.cache
+# old php-cs-fixer v2
 .php_cs.cache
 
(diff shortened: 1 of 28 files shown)


phil-davis:
I have done this PR to demonstrate that this code-style update is easy to do. It creates "quite a few" lines of diff.
If people are K with applying this in the oC10 app repos, then I can tell people here that they can do a few in between other jobs, when waiting for a CI result, or tasks are otherwise blocked.

JammingBen (approved):
LGTM 🚀

PR #959, opened 2021-06-13
[tests-only] Allow for no unit tests with coverage in CI

phil-davis:
Problem: Some JS and/or PHP test pipelines do not have coverage enabled. If there are no pipelines with coverage enabled, the sonar-analysis pipeline is still being added to CI. But it has nothing to do, and fails because it does not find any coverage results to process.

Solution: this PR refactors `.drone.star` so that it:
1) finds all JS and PHP test pipelines that do have coverage enabled
2) adds the sonar-analysis pipeline if (1) is non-empty
3) adds any JS and PHP test pipelines that do not have coverage enabled
4) continues adding the acceptance test pipelines.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -144,6 +144,13 @@ def main(ctx):
 
 	dependsOn(before, coverageTests)
 
+	nonCoverageTests = nonCoveragePipelines(ctx)
+	if (nonCoverageTests == False):
+		print('Errors detected in nonCoveragePipelines. Review messages above.')
+		return []
+
+	dependsOn(before, nonCoverageTests)
+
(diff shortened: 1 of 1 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/activity/1933
The pipelines with coverage appear before the sonar-analysis pipeline.
The unit test pipelines without coverage appear after the sonar-analysis pipeline.
That is what I expect.

See the gallery app PR https://github.com/owncloud/gallery/pull/842 for an example of this code when there are no pipelines that have coverage - in that case there is no sonar-analysis pipeline, which is what we want.

individual-it (approved):

Issue #965, opened 2021-06-21
[QA] Expired public share has wrong activity message


JammingBen:
### Steps to reproduce
1. Create a public link
2. Let it expire
3. Check your activity stream

### Expected behaviour
It should say "Your public link for xyz has expired."

### Actual behaviour
It says "You removed the public link for xyz"

### Server configuration
**Operating system**: Mac OS

**Web server:** Apache

**Database:** SQLite

**PHP version:** 7.4

**ownCloud version:** OC 10.8.0-RC1

PR #967, opened 2021-06-21
Fix check for expired public links

JammingBen:
fixes https://github.com/owncloud/activity/issues/965

### DIFF
--- a/lib/FilesHooks.php
+++ b/lib/FilesHooks.php
@@ -436,7 +436,7 @@ class FilesHooks {
 		if ($isSharing) {
 			$actionSharer = 'shared_link_self';
 			$actionOwner = 'reshared_link_by';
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: there have not been any automated tests for what happens after a link or share expires, because it was difficult to "wait until tomorrow"  in CI for a link or share to reach its expiry.

But now we have a step in core:
`When the administrator expires the last created share using the testing API`
or
`Given the administrator has expired the last created share using the testing API`

That is supported by the testing app, which sets the expiration date of the share to yesterday. Then tests can access the share, and the server automatically detects that the share expires and removes it.

We could use that to add automated tests for share expiration messages in the activity app. I created issue #968 for that.

JanAckermann (approved):

Issue #970, opened 2021-06-21
[QA] Activity when receiving a federated share is missing in sidebar and `Activities by others`


JammingBen:
### Steps to reproduce
1. Receive and accept a share from a federated OC system
2. Check the sidebar and you activity stream

### Expected behaviour
The activity should be listed in the file sidebar and in the activity stream in `Activities by others`

### Actual behaviour
The activity is missing in the file sidebar as well as in `Activities by others`

JanAckermann:
As in #623

Issue #969, opened 2021-06-21
[QA] Download activity is missing in `Activities by others`


JammingBen:
### Steps to reproduce
1. Create a public link share
2. Open the public link in another browser and download the file
3. View activity stream

### Expected behaviour
The activity should be listed in `Activities by others`

### Actual behaviour
The activity is missing in `Activities by others`

PR #972, opened 2021-06-23
Set the objectId for a received federated share activity

JammingBen:
Before this fix, the objectId of a received federated share activity was empty. As a result, the activity was not showing in the file sidebar. We now set the objectId by updating the activity as soon as the share has been accepted by the user.

Also fixed a bug where certain activities were not showing in `Activities by others`.

- fixes https://github.com/owncloud/activity/issues/969
- fixes https://github.com/owncloud/activity/issues/970

- Depends on https://github.com/owncloud/core/pull/38880

image

### DIFF
--- a/lib/Hooks.php
+++ b/lib/Hooks.php
@@ -79,6 +79,7 @@ class Hooks {
 	public static function onRemoteShareAccepted($params) {
 		$shareId = $params->getArgument('shareId');
 		$fileId = $params->getArgument('fileId');
+		$shareRecipient = $params->getArgument('shareRecipient');
 
 		if ($shareId === null || $fileId === null) {
 			return;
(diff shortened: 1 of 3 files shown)


JanAckermann (changes requested):
Waiting until owncloud/core#38880 gets merged

lib/Hooks.php
```diff
@@ -67,4 +70,32 @@ protected static function deleteUserMailQueue(IDBConnection $connection, $user)
...
+		$queryBuilder = $connection->getQueryBuilder();
+
+		$queryBuilder->update('activity')
+			->set('object_id', $queryBuilder->createNamedParameter($fileId))
+			->where($queryBuilder->expr()->eq('subject', $queryBuilder->createParameter('subject')))
+			->andWhere(
+				$queryBuilder->expr()->like('subjectparams', $queryBuilder->createParameter('subjectparams'))
+			)
+			->setParameter('subject', Activity::SUBJECT_REMOTE_SHARE_RECEIVED)
+			->setParameter('subjectparams', '%{"shareId":"' . $shareId . '"}%');
```
jvillafanez, line 97: Check for alternatives because this might perform quite bad. There is no index in any of those columns, and the activity table might grow very large.

JanAckermann: We could add the affected user to the where query? I don't see alternatives otherwise.

JammingBen: We added some more params and ran an additional `explain` statement: `explain select * from oc_activity WHERE affecteduser = 'admin' and subjectparams LIKE '%test%' and object_id = 0 and subject = 'remote_share_received';`

Which seems to result in a index scan:

```
Index Scan using activity_filter_app on oc_activity  (cost=0.14..8.16 rows=1 width=3992)
  Index Cond: ((affecteduser)::text = 'admin'::text)
  Filter: ((subjectparams ~~ '%test%'::text) AND (object_id = 0) AND ((subject)::text = 'remote_share_received'::text))
```

jvillafanez (approved):

JanAckermann (approved):

PR #978, opened 2021-06-29
[tests-only] make code coverage report correctly on PRs

phil-davis:
Issue https://github.com/owncloud/QA/issues/678

1st commit) Adjust sonar-analysis step so that it overrides the default "clone" step of drone, like core PR https://github.com/owncloud/core/pull/38842

2nd commit) refactor so that we do not need `config['app']` - we can use `ctx.repo.name` to get the current name of the "app".

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1,5 +1,4 @@
 config = {
-	'app': 'activity',
 	'rocketchat': {
 		'channel': 'builds',
 		'from_secret': 'private_rocketchat'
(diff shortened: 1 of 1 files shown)


phil-davis:
The analysis results are at https://sonarcloud.io/dashboard?id=owncloud_activity&pullRequest=978

Note: we won't see the coverage reported until this PR is merged, and other PRs happen that actually have code changes where coverage diffs for the PR  can be calculated. (This PR has 0 lines of "real" code changed, of which 0 lines are covered by tests, and 0/0 = ?)

individual-it (approved):

PR #979, opened 2021-06-29
[tests-only] Add some rubbish code and test

phil-davis:
This is just to demonstrate that the SonarCloud coverage report for the PR is now working.


phil-davis:
That's good - it gave me 75% coverage of the changes.

PR #991, opened 2021-07-28
[tests-only] Format starlark

phil-davis:
1) replace tab-indent with space-indent in `.drone.star`
2) add `check-starlark` pipeline to check the formatting of the drone starlark code - the same as in `owncloud/web`

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -171,7 +171,7 @@ def main(ctx):
     return before + coverageTests + afterCoverageTests + nonCoverageTests + stages + after
 
 def beforePipelines(ctx):
-    return codestyle(ctx) + jscodestyle(ctx) + phpstan(ctx) + phan(ctx)
+    return codestyle(ctx) + jscodestyle(ctx) + phpstan(ctx) + phan(ctx) + checkStarlark()
 
 def coveragePipelines(ctx):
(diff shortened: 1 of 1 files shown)


phil-davis:
This makes the activity app have similar format `.drone.star` like what was done in https://github.com/owncloud/web/pull/4987

If we do this, then next time we have `.drone.star` changes to update through all oC10 apps I will apply this formatting change.

Is this "a good thing" ?

wkloucek (approved):
:+1:

wkloucek:
> This makes the activity app have similar format `.drone.star` like what was done in [owncloud/web#4987](https://github.com/owncloud/web/pull/4987) and [owncloud/ocis#2023](https://github.com/owncloud/ocis/pull/2023)
> 
> If we do this, then next time we have `.drone.star` changes to update through all oC10 apps I will apply this formatting change.
> 
> Is this "a good thing" ?

First of all it's hard work to apply this on all pipelines. I have been annoyed that I can't copy functions between files and then just apply a formatter because then the whole diff is not only reflecting my changes... I would say it's a good thing to enforce a certain format like we do it for all other code, too

xoxys (approved):

PR #1000, opened 2021-08-23
[tests-only] Apply changes to .drone.star from search_elastic app

phil-davis:
1) These changes happened in https://github.com/owncloud/search_elastic/pull/215

Apply them here so that we have the starlark up-to-date.

2) Make a fix to handle the `buildJsDeps` checks better.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1704,6 +1704,11 @@ def installApp(ctx, phpVersion):
     if "appInstallCommand" not in config:
         return []
 
+    if "buildJsDeps" not in config:
+        installJsDeps = False
+    else:
+        installJsDeps = config["buildJsDeps"]
+
     return [
         {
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -1714,7 +1720,7 @@ def installApp(ctx, phpVersion):
                 "make build-dev",
             ],
         },
-    ] if config["buildJsDeps"] else [] + [{
```
phil-davis, line 1717: Note: I got an error here when applying this in the search_elastic app. The starlark crashed if `buildJsDeps` was not set in the `config` array. So I adjusted this code. It works now in https://github.com/owncloud/search_elastic/pull/226

dpakach (approved):
LGTM 👍🏼

PR #1018, opened 2021-10-27
Implement activities for rename and move actions

JammingBen:
Needs https://github.com/owncloud/core/pull/39430

For acceptance test scenarios see https://github.com/owncloud/activity/issues/1019

- fixes https://github.com/owncloud/enterprise/issues/4806
- fixes https://github.com/owncloud/activity/issues/375
- fixes https://github.com/owncloud/enterprise/issues/4375
- fixes https://github.com/owncloud/enterprise/issues/2947


phil-davis (approved):
LGTM

lib/FilesHooks.php
```diff
@@ -134,15 +137,59 @@ public function fileRestore($path) {
 		$this->addNotificationsForFileAction($path, Files::TYPE_SHARE_RESTORED, 'restored_self', 'restored_by');
 	}
 
+	/**
+	 * After rename/move events
+	 * @param string $oldPath Path of the file before rename
+	 * @param string $newPath Path of the file after rename
+	 */
+	public function fileAfterRename($oldPath, $newPath) {
+		// rename
```
jvillafanez, line 146: We might need a comment here to say that the .part files and the missing files (the 2 cases handled also in the `fileBeforeRename` method) will be handled in the `addNotificationsForFileAction` method.

lib/FilesHooks.php
```diff
@@ -154,38 +201,80 @@ protected function addNotificationsForFileAction($filePath, $activityType, $subj
...
+			}
+
+			if ($activityType === Files::TYPE_FILE_MOVED) {
+				$oldUserPath = $oldAffectedUsers[$user] ?? null;
+				$newUserPath = $newAffectedUsers[$user] ?? null;
+
+				if ($oldUserPath && $newUserPath) {
+					// User has old and new path -> regular move action. Add the old path as additional info.
+					$userParams[] = $oldUserPath;
+				} elseif (!$newUserPath) {
```
jvillafanez, line 245: `$newUserPath === null` might be better to prevent issues with `"0"` being false.

lib/FilesHooks.php
```diff
@@ -154,38 +201,80 @@ protected function addNotificationsForFileAction($filePath, $activityType, $subj
...
+					// No new path -> file was moved somewhere the user has no access to (e.g. out of a share).
+					$userSubject = 'deleted_by';
+					$computedActivityType = Files::TYPE_SHARE_DELETED;
+				} else {
+					// No old path -> file was moved from somewhere the user has no access to (e.g. into a share).
```
jvillafanez, line 250: It isn't clear whether the case where both newUserPath and oldUserPath is handled properly or not. If that case shouldn't happen, it might be better to add a comment

lib/FilesHooks.php
```diff
@@ -154,38 +201,80 @@ protected function addNotificationsForFileAction($filePath, $activityType, $subj
...
+			}
+
+			if ($activityType === Files::TYPE_FILE_RENAMED) {
+				$userParams[] = $oldAffectedUsers[$user] ?? $oldPath;
+
+				// Share itself gets renamed -> it only affects the user who renamed.
+				// For other users, the old and new path stay the same -> no activity.
+				$oldUserPath = $oldAffectedUsers[$user] ?? null;
+				$newUserPath = $newAffectedUsers[$user] ?? null;
+				if ($oldUserPath && $newUserPath && $oldUserPath === $newUserPath) {
```
jvillafanez, line 263: should we notify if both oldUserPath and newUserPath are null?

lib/FilesHooksStatic.php
```diff
@@ -29,12 +29,19 @@
...
+	private static $app = null;
+
 	/**
 	 * @return FilesHooks
 	 */
 	protected static function getHooks() {
-		$app = new AppInfo\Application();
-		return $app->getContainer()->query('Hooks');
+		if (self::$app === null) {
+			self::$app = new AppInfo\Application();
```
jvillafanez, line 42: It might be better if we add a `public static function setHooks(...)`. The initialization would happen in the `Application` class.

lib/FilesHooks.php
JammingBen: I added a comment which explains that a scenario where `$oldUserPath` and `$newUserPath` both are null is technically not possible. Same goes for the rename case down below.

lib/FilesHooksStatic.php
JammingBen: The `FileHookStatic` class is never being initialized, see https://github.com/owncloud/activity/blob/master/lib/AppInfo/Application.php#L313.

If we initialize the `FilesHooksStatic` class, then we could just use the `FilesHooks` class and make the static class obsolete. While I think this would be a good thing (if possible), I wouldn't make it part of this PR.

jvillafanez:
```
public function registerFilesActivity() {
  FilesHooksStatic::setHooks($this->getContainer()->query('Hooks'));
  // All other events from other apps have to be send via the Consumer
  Util::connectHook('OC_Filesystem', 'post_create', 'OCA\Activity\FilesHooksStatic', 'fileCreate');
  ....
}
```
That should be enough I think. We might include a log if possible in case we overwrite it with a new instance.

JammingBen: Updated the PR, it seems to cause problems now though: https://drone.owncloud.com/owncloud/activity/2302/4/2

ownclouders:
:boom: Acceptance tests pipeline <strong>apiActivity-master-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/activity/2302/25/1

lib/FilesHooksStatic.php
jvillafanez: We might need to provide a fallback. I mean, normally the app should be properly initialized, so the `registerFilesActivity` is called and the Hooks are injected in the `FilesHooksStatic` static class. In case this doesn't happen, we might need to get the Hooks from a new application instance

JammingBen: Doesn't fix it, unfortunately. I guess it still has something to do with calling `FilesHooksStatic::setHooks()` in the application class. Or do you see any other issue?

JammingBen: I reverted the change for now as `$this->getContainer()->query('Hooks')` won't work in the `registerHooks()` method. It would try to fetch the current user, which is not (yet) available during the installation process.

IMO the current solution is not too bad. Also, it needs to/should be be rewritten anyway once we switch to the Symfony event dispatcher.

jvillafanez (approved):

phil-davis:
@JammingBen IMO this can be merged.

@kiranparajuli589 has an idea for a couple of public-link rename/move test scenarios - but he will add those in a later PR.

PR #1031, opened 2021-11-16
Implement a global setting to enable or disable move/rename-activities

JammingBen:
This is an addition to https://github.com/owncloud/activity/pull/1018.

Per default, move and rename activities are disabled. To enable run: `php occ config:app:set activity enable_move_and_rename_activities --value yes`. When disabling after this setting was enabled before, rename/move activities from the past will still show up in the activity stream.


ownclouders:
:boom: Acceptance tests pipeline <strong>apiActivity-master-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/activity/2320/25/1

JammingBen:
@phil-davis I added acceptance tests to check that there is no activity when the setting is disabled. I noticed that `Then user "Alice" should not have any activity entries with type "file_renamed$"` always succeeds, regardless of the setting. Any idea why?

phil-davis:
> Any idea why?

Sorry, I should have noticed straight away. The "$" at the end means that there are no activity entries with the literal type `files_renamed$`

I tried locally with `file_created` to make sure that I could make the step fail:
```
    Then user "Alice" should not have any activity entries with type "file_created"                                  # ActivityContext::userShouldNotHaveAnyActivityEntriesWithType()
      found an activity entry of type file_created for user Alice which should not exist
      Failed asserting that 'file_created' is not equal to 'file_created'.
```

I pushed 2 commits:
1) removed the `$` from the activity type strings.
2) enhanced the output when `Then user :user should not have any activity entries` fails, so that it reports the activity types that were found - that makes it easy to paste in a step like `Then user "Alice" should not have any activity entries` and the error message when it fails will show which activity types were unexpectedly found.

JanAckermann (approved):

PR #1040, opened 2021-12-21
[tests-only] Trigger CI

phil-davis:
To check if it works with the latest `owncloudci/core` docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 Activity App
 ===============
 
+Trigger CI
+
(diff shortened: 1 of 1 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/activity/2393/13/5
`latest: Pulling from owncloudci/core`

that installed OK. It does not need to do `composer` or `yarn` commands because in this case all dependencies are already in the core tarball.

PR #1071, opened 2022-02-27
Discard activity emails for users with invalid email addresses

mdusher:
This adds email address validation to the background job for the event a user has been able to set an invalid email address on their account.

It will discard the email queue for the user if they have an invalid email address to avoid a backlog of items being queued up.


mdusher:
Just going to close this until I have time to come back to it.

PR #1080, opened 2022-03-14
Test touch php unit

phil-davis:


phil-davis:
All unit tests ran (PHP and JS) - that's fine, we are not being that fine-grained to have it understand if only the unit tests of one language are changed.

Acceptance tests were skipped - correct.

PR #1074, opened 2022-03-10
[CI] Implement skip-if-unchanged in drone

SwikritiT:
part of : 
https://github.com/owncloud/QA/issues/730

PR #1086, opened 2022-03-14
Test touch composer

phil-davis:
All tests need to run.


phil-davis:
tests were skipped - that is a problem. I will adjust in PR #1074

PR #1094, opened 2022-03-21
Enable files external in federated server

phil-davis:


phil-davis:
still fails. IMO there is a core test issue. It look like the scenarios are not running in core, and have a problem there.
I am working on it in core PR https://github.com/owncloud/core/pull/39902

PR #1098, opened 2022-04-01
[tests-only][full-ci] compare full ci runs

kiranparajuli589:
Signed-off-by: Parajuli Kiran <kiranparajuli589@gmail.com>

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -50,7 +50,9 @@ class Application extends App {
 		$container = $this->getContainer();
 
 		/**
-		 * Activity Services
+				 * Activity Services
+				 * returns new data with activity manager
+				 *
 		 */
 		$container->registerService('ActivityData', function (IContainer $c) {
 			/** @var \OC\Server $server */


kiranparajuli589:
build finished taking 29m 19s time https://drone.owncloud.com/owncloud/activity/2655
before drone enhancement PRs, https://drone.owncloud.com/owncloud/activity/2622 took 22m 1s :) it was shorter before.
but a number of other past full ci runs took a similar time range of finishing at 25 min to 30 min like https://drone.owncloud.com/owncloud/activity/2613

IMO, after the changes on behalf of https://github.com/owncloud/QA/issues/732:
- implemented cancel previous build system so that there are no duplicated builds for the same pr. this will also save resources.
- skipping un-related pipelines has been enhanced (see more details https://github.com/owncloud/QA/issues/730).
- use of variables makes developing on starlark easier
- outdated and not maintained images use cases are removed and replaced with well-maintained images
- using chcace images for pipelines should save us some time but it is not making that much of difference. maybe some minutes in some runs.

PR #1108, opened 2022-05-18
[CI][full-ci]Move unit tests into tests/unit folder for activity app

SagarGi:
This PR moves unit tests into tests/unit folder. And also tidy up the `phpunit.xml` to make standard format for all oc-apps as much as possible.

- Part of https://github.com/owncloud/impersonate/issues/198


composer.lock
```diff
@@ -68,5 +68,5 @@
     "platform-overrides": {
         "php": "7.3"
     },
-    "plugin-api-version": "2.2.0"
+    "plugin-api-version": "1.1.0"
```
phil-davis, line 71: @SagarGi I currently have Composer 2.3.5 on my system, and `composer update` puts an up-to-date plugin-api-version in `composer.lock` for me. I don't remember if there is also a separate composer plugin that also needs to be updated.

Anyway, this diff should not be needed.

tests/bootstrap.php
```diff
@@ -0,0 +1,14 @@
...
+
+require_once __DIR__ . '/../../../lib/base.php';
+
+// especially with code coverage it will require some more time
+\set_time_limit(0);
+
+\OC::$composerAutoloader->addPsr4('Test\\', OC::$SERVERROOT . '/tests/lib/', true);
+\OC::$composerAutoloader->addPsr4('TestHelpers\\', OC::$SERVERROOT . '/tests/TestHelpers/', true);
+// load activity unit test classes
+\OC::$composerAutoloader->addPsr4('OCA\Activity\Tests\Unit\\', __DIR__ . '/unit/', true);
```
saw-jan, line 14: We have two options here:
1. Change namespace from `OCA\Activity\Tests\Unit` to `OCA\Activity\Tests\unit;` in all test files.
2. Create separate bootstrap file (**done in this PR**)

SagarGi: @phil-davis some apps have cases like this. You can have a look to it and can be followed accordingly for other apps too.

phil-davis (approved):
LGTM. I removed a few lines from `tests/unit/bootstrap.php` that were not needed.

PR #1118, opened 2022-07-05
Remove the time from the mail, keep the date

jvillafanez:
Related to https://github.com/owncloud/enterprise/issues/5227

The specific time of the event won't be shown, just the date. Note that "Today", "Yesterday" or "Tomorrow" might still appear instead of the actual date.


phil-davis (approved):

jnweiger:
Confirmed fixed. The time is actually removed from *all* notification mails, not only expiry mails. 
That is okayish, as the mail itself still has a timestamp in the header.

```
"Date": "Mon, 12 Sep 2022 21:39:10 +0000",
"To": "admin <admin@example.com>",
"Subj": "Activity notification",

Hello admin,

You are receiving this email because the following things happened at https://oc10110rc1-activity-271rc1-20220912.jw-qa.owncloud.works/

* You shared Documents with group admin - Today
* You created Photos/bunny.stl - Today

```

PR #1138, opened 2022-10-07
[tests-only] [full-ci] Run CI with PHP 7.4-ubuntu20.04

phil-davis:
This is a test to see if it passes. We can decide next week if we explicitly change to `owncloudci/php:7.4-ubuntu20.04` or if we update `owncloudci/php:7.4` so that it uses Ubuntu 20.04 (instead of Ubuntu 18.04) or if we design some other naming system for tagging the Docker images so that we know which Ubuntu+PHP combinations are being used.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1130,12 +1130,19 @@ def acceptance(ctx):
                     continue
 
                 name = "unknown"
+                phpVersionForDocker = testConfig["phpVersion"]
+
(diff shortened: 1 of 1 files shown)


phil-davis:
CI passes - good.

phil-davis:
See #1139 - we do not need to actually change the docker image tag any more.`owncloudci/php:7.4` was updated so that it is now based on Ubuntu 20.04

PR #1139, opened 2022-10-13
[tests-only] [full-ci] Adjust drone code to allow for variation of PHP docker images

phil-davis:
This is code from PR #1138 without actually changing the `owncloudci/php` docker image. It adds code to just use the first 3 characters of the `owncloudci/php` docker image tag when generating pipeline names and coverage report names.

Similar code was done in wopi https://github.com/owncloud/wopi/pull/220 and core https://github.com/owncloud/core/pull/40409

It will be useful to have this merged here in the activity app. Then when we next need to propagate CI code changes to other apps, this will be here.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -851,6 +851,12 @@ def phpTests(ctx, testType, withCoverage):
             else:
                 command = "make test-php-integration"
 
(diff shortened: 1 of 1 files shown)


saw-jan (approved):
LGTM

individual-it (approved):

PR #1142, opened 2022-10-20
[tests-only] [full-ci] Use browserService in more places in .drone.star

phil-davis:
These were in `.drone.star` in brute_force_protection app. Do the same here in the activity app.


phil-davis:
As I review the PRs in other oC10 app repos, I will push this also as needed.

PR #1155, opened 2022-11-22
[tests-only] Check PHP code with 7.3

phil-davis:
oC10 core up to 10.11 supports PHP 7.3. For the time being, when we release core apps, we want to know that the PHP code works OK with PHP 7.3. So run code-style in CI with PHP 7.3

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -45,7 +45,14 @@ config = {
     "branches": [
         "master",
     ],
-    "codestyle": True,
+    "codestyle": {
+        "multiple": {
+            "phpVersions": [
+                DEFAULT_PHP_VERSION,
+                "7.3",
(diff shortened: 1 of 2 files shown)


phil-davis:
`owncloud/coding-standard` needs to use major version 3 to work with PHP 7.3.
But there are some code-style differences between version 3 and version 4. So the pipeline for `owncloud/coding-standard` v3 with PHP 7.3 fails due to some minor indent differences.

That's annoying - I really only cared about the "PHP lint" part of php-cs-fixer passing.

I will think about it tomorrow - what is the best way to get a "lint" (syntax check) with PHP 7.3

phil-davis:
Done differently in #1156

PR #1157, opened 2022-12-01
[tests-only] [full-ci] Add wait-for-samba drone CI code

phil-davis:
This is the new code from WND PRs https://github.com/owncloud/windows_network_drive/pull/418 and https://github.com/owncloud/windows_network_drive/pull/419

Adding it here in the activity app CI logic so that it will be in the "standard" oC10 app CI code, and will get copied to the other repos next time we need to do a general update of the drone starlark rate.


phil-davis:
The wait-for-samba pipeline step was not used in the activity app CI - good. The logic worked - there is no samba service to wait-for.

SwikritiT (approved):

PR #1159, opened 2022-12-02
Bump phpstan version to 1.9

amrita-shrestha:
### Description
- use the latest phpstan, which is 1.9.* series

### Related issue
- https://github.com/owncloud/QA/issues/777

### DIFF
--- a/vendor-bin/phpstan/composer.json
+++ b/vendor-bin/phpstan/composer.json
@@ -1,5 +1,5 @@
 {
     "require": {
-        "phpstan/phpstan": "^1.8"
+        "phpstan/phpstan": "^1.9"
     }
 }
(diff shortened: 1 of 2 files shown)


vendor-bin/phpstan/composer.json
```diff
@@ -1,5 +1,5 @@
 {
     "require": {
-        "phpstan/phpstan": "^1.8"
+        "phpstan/phpstan": "^1.9"
```
phil-davis, line 3: Note: `^1.8` means use any `1.*` version from 1.8 onward. So in CI the latest 1.9.2 is already running:
https://drone.owncloud.com/owncloud/activity/3090/3/4

But it does not hurt to update here in `composer.json` so that we know that at least 1.9 is really being used.

phil-davis (approved):

Issue #1176, opened 2023-05-22
Release activity 2.7.2


phil-davis:
### Reasons

- Needed to go with server core 10.13 because of the Symfony 4 to 5 upgrade https://github.com/owncloud/core/issues/39630
- PR #1173 made code changes needed for Symfony 5
- #1181 (not yet merged)??

Note: the change in that PR also works fine with server core 10.12 and before, so there should be no need to bump the core min-version.

The app can be released any time before 10.13 - it should work fine with any of the recent oC10 core releases.

## Needs full QA
- Testplan #1184

jnweiger:
Release done together with 10.13.0

PR #1181, opened 2023-07-04
Readd time to mail for non-expiry activities

pako81:
In https://github.com/owncloud/activity/pull/1118 we removed the time from the notification emails because of the confusion generated by the activity time for shares expiry.

However, this caused the timestamp to be removed for all activity notifications. This PR adds logic for having the correct date format depending on the activity type (expiration vs non-expiration).

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1,12 +1,10 @@
 BANST_AWS_CLI = "banst/awscli"
-DRONE_CLI = "drone/cli:alpine"
 INBUCKET_INBUCKET = "inbucket/inbucket"
 MINIO_MC = "minio/mc:RELEASE.2020-12-18T10-53-53Z"
 OC_CI_ALPINE = "owncloudci/alpine:latest"
 OC_CI_BAZEL_BUILDIFIER = "owncloudci/bazel-buildifier"
 OC_CI_CEPH = "owncloudci/ceph:tag-build-master-jewel-ubuntu-16.04"
 OC_CI_CORE = "owncloudci/core"
-OC_CI_DRONE_CANCEL_PREVIOUS_BUILDS = "owncloudci/drone-cancel-previous-builds"
 OC_CI_DRONE_SKIP_PIPELINE = "owncloudci/drone-skip-pipeline"
 OC_CI_NODEJS = "owncloudci/nodejs:%s"
 OC_CI_ORACLE_XE = "owncloudci/oracle-xe:latest"
@@ -207,7 +205,7 @@ def main(ctx):
     return before + coverageTests + afterCoverageTests + nonCoverageTests + stages + after
 
 def beforePipelines(ctx):
-    return validateDailyTarballBuild() + codestyle(ctx) + jscodestyle(ctx) + cancelPreviousBuilds() + phpstan(ctx) + phan(ctx) + phplint(ctx) + checkStarlark()
+    return validateDailyTarballBuild() + codestyle(ctx) + jscodestyle(ctx) + phpstan(ctx) + phan(ctx) + phplint(ctx) + checkStarlark()
 
 def coveragePipelines(ctx):
     # All unit test pipelines that have coverage or other test analysis reported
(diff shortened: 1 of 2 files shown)


phil-davis:
I added a commit to remove some special drone actions from CI - we are cleaning up various things that need special privileges/tokens and the needed tokens no longer exist. Since this PR has just been created and is failing because of this, I may as well fix it here and get a good CI result.

phil-davis:
https://drone.owncloud.com/owncloud/activity/3427/6/7
The workflow that gathers unit test coverage pushes the individual coverage results into an S3 cache and the SonarCloud coverage analysis runs with the data from that cache. The cache token(s) are in the process of being sorted out. so the unit test coverage fails for now.

jvillafanez:
I'm not sure about the implementation.
If the idea is to target one or two specific events, you could do something like `if ($activity['amq_subjectparams'] === 'shareExpired') {....}` (not sure if we can use a constant)
While it can be hard to find an event that contains "shareExpired" which we don't want to format appropriately, it could "accidentally" happen. I mean, false positives could happen.

In order to target a group of events, you'll have to distinguish those events somehow. I guess we could use a `$activity['eventType'] === 'expiringEvent'` or even better `$activity['dateFormat'] === 'formatDateRelativeDay'` so it's the actual activity's sender the one deciding the formatting.

pako81:
Yes, I also thought about targeting a group of events but did not find a reliable way. `getItemsForUser` is used for getting all items for the user we want to send an email to from the `oc_activity_mq` table https://github.com/owncloud/activity/blob/master/lib/MailQueueHandler.php#L147 and this is how the table looks like:

```
+-------------------+---------------+------+-----+---------+----------------+
| Field             | Type          | Null | Key | Default | Extra          |
+-------------------+---------------+------+-----+---------+----------------+
| mail_id           | bigint(20)    | NO   | PRI | NULL    | auto_increment |
| amq_timestamp     | int(11)       | NO   | MUL | 0       |                |
| amq_latest_send   | int(11)       | NO   | MUL | 0       |                |
| amq_type          | varchar(255)  | NO   |     | NULL    |                |
| amq_affecteduser  | varchar(64)   | NO   | MUL | NULL    |                |
| amq_appid         | varchar(255)  | NO   |     | NULL    |                |
| amq_subject       | varchar(255)  | NO   |     | NULL    |                |
| amq_subjectparams | varchar(4000) | NO   |     | NULL    |                |
+-------------------+---------------+------+-----+---------+----------------+
```

So what we get at https://github.com/owncloud/activity/blob/master/lib/MailQueueHandler.php#L253 is just an array of those fields. So the only valid way for filtering for expiry events seemed to use the `amq_subjectparams` field.

jnweiger:
Good to see progess here!
I'd like to include this for the 10.13.0 release - see #1176 - please retarget this PR into the release-2.7.2 branch if you agree.

pako81:
Just noticed that public link expiration contains the following in `amq_subjectparams`:

```
[{"1715":"\/public"},"auto:automation"]
```

while expiration for users/group shares has:

```
[{"1714":"\/New folder"},"auto:automation","shareExpired"]
```

so the PR is in any case incomplete as with the code in place we are missing the case for public links expiration.

Maybe we can try to `strpos` the string `automation` which is included in both cases or an alternative approach would be to use `amq_subject` in the `if` condition. Users/group share expiration has `unshared_by` while public links have `link_expired`. However, note that `unshared_by` is also used for the share recipient when the sharer manually unshare the resource. So it will not be unique to the expiry event only.

jvillafanez:
The `amq_subjectparams` is json-encoded (https://github.com/owncloud/activity/blob/ce947de1ad5cd36fb75c3cb295d2890ebbb8b111/lib/Data.php#L158) so we should treat it as json, not a plain string.

What about including new column for options, or something like that? We could have a column with a bit mask, for example:
* 1 -> use datetime instead of date for the amq_timestamp
* 2 -> format short date instead of long date
* 4 -> format short time instead of long time
Assuming we use a 32-bit integer, we still have margin to add additional options later.

For the migration, we can assume that all the current data will have a 0 value as option, which means there won't be changes.

pako81:
@jvillafanez it looks like the activity file hooks already include the info about the share being or not expired: https://github.com/owncloud/activity/blob/master/lib/FilesHooks.php#L358

What about adding a new column `expiry` which could have the value `0` and `1` depending if we are dealing or not with an expire event and then use date or datetime based on the value of the column.

jvillafanez:
I think that's more or less the idea I've proposed.

Taking into account that we might not be allowed to add new columns into the tables freely, I think we should try to make it count.
Adding the `expire` column is fine, but I think it only fits for this particular use case. For any other similar use case we'll need to make more changes and likely add new columns or tables. I'd prefer to have something that could be used in more use cases even if we only implement this one right now.

PR #1182, opened 2023-07-04
[CI] [full-ci] Remove special PR-based drone actions from CI

phil-davis:
- remove the use of `drone_token`
- adjust the use of the `cache_s3` tokens to use the names now available

Part of issue https://github.com/owncloud/QA/issues/818

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -26,6 +26,22 @@ THEGEEKLAB_DRONE_GITHUB_COMMENT = "thegeeklab/drone-github-comment:1"
 DEFAULT_PHP_VERSION = "7.4"
 DEFAULT_NODEJS_VERSION = "14"
 
+# minio mc environment variables
+MINIO_MC_ENV = {
+    "CACHE_BUCKET": {
+        "from_secret": "cache_s3_bucket",
+    },
+    "MC_HOST": {
+        "from_secret": "cache_s3_server",
+    },
+    "AWS_ACCESS_KEY_ID": {
+        "from_secret": "cache_s3_access_key",
+    },
+    "AWS_SECRET_ACCESS_KEY": {
+        "from_secret": "cache_s3_secret_key",
+    },
+}
+
 dir = {
     "base": "/var/www/owncloud",
     "federated": "/var/www/owncloud/federated",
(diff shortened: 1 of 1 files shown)


phil-davis:
I can do this locally:
Define 3 environment variables:
```
$ export ACCESS_KEY="access123"
$ export SECRET_KEY="secret456"
$ export HOSTURL="https://somewhere.something.com"
```

Now construct the URL format like:
```
$ export URL="https://$ACCESS_KEY:$SECRET_KEY@${HOSTURL#https://}"; echo $URL
https://access123:secret456@somewhere.something.com
```

That works fine locally for me in a terminal. But when I use that sort of thing in a command in a drone pipeline:
```
                             "export MC_HOST_cache=\"https://$S3_ACCESS_KEY:$S3_SECRET_KEY@${S3_SERVER#https://}\"; mc mirror cache/cache/%s/%s results/" % (ctx.repo.slug, ctx.build.commit + "-${DRONE_BUILD_NUMBER}"),
```

I get drone output https://drone.owncloud.com/owncloud/activity/3444/10/5
```
+ export MC_HOST_cache="https://$S3_ACCESS_KEY:$S3_SECRET_KEY@"; mc mirror cache/cache/owncloud/activity/ca77c1fd080c895654792b93bfa158d22df7f3c4-3444 results/

mc: <ERROR> Unable to stat source `cache/cache/owncloud/activity/ca77c1fd080c895654792b93bfa158d22df7f3c4-3444`. Endpoint:  does not follow ip address or domain name standards.
```

.drone.star
```diff
@@ -1351,13 +1324,19 @@ def sonarAnalysis(ctx, phpVersion = DEFAULT_PHP_VERSION):
...
+                                 "from_secret": "cache_s3_secret_key",
+                             },
+                             "S3_SERVER": {
+                                 "from_secret": "cache_s3_server",
                              },
                          },
                          "commands": [
                              "mkdir -p results",
-                             "mc mirror cache/cache/%s/%s results/" % (ctx.repo.slug, ctx.build.commit + "-${DRONE_BUILD_NUMBER}"),
+                             "export MC_HOST_cache=\"https://$S3_ACCESS_KEY:$S3_SECRET_KEY@${S3_SERVER#https://}\"; mc mirror cache/cache/%s/%s results/" % (ctx.repo.slug, ctx.build.commit + "-${DRONE_BUILD_NUMBER}"),
```
phil-davis, line 1339: I want the `mc mirror` command here to have the correct value for env var `MC_HOST_cache` like it did before.
But now I need to construct it from `cache_s3_access_key` `cache_s3_secret_key` and `cache_s3_server`
And the `cache_s3_server` value needs to have the `https://` stripped off it (and put at the very beginning of the whole string.

But the `${S3_SERVER#https://}` part does not seem to work.

What have I done wrong?

phil-davis: `${S3_SERVER:8}` also does not work (trying to get the string from position 8, ignoring "http://")
Maybe drone is smart and refuses to do things that "chop up" the string value, because that would make it difficult for drone to keep track of where parts of the secret are stored, and to protect from every echo/displaying the value anywhere.

phil-davis: I found https://0-8-0.docs.drone.io/manage-secrets/

"Please note parameter expressions are subject to pre-processing. When using secrets in parameter expressions they should be escaped."

`$${S3_SERVER:8}` works.

phil-davis:
@micbar @xoxys I found the tricks for being able to construct the `MC_HOST_cache` value.

I think that this will get CI passing again.

Please review - if it looks OK then I will use this as the pattern in other repos that have this kind of drone starlark.

After this, this will still be the GitHub token issue - if a PR build feels the need to post a comment to GitHub then it will not be able to do that. That will be OK for today/tomorrow - the pipeline is failing anyway!

Then we can discuss if there is some way to have a "github comment-only token" that can be made available to PRs, or if we remove the comment-in-PR feature.

I am OOTO for a few hours...

xoxys:
I still think the mc alias approach should be used. You have to touch all repos anyway, why not use the IMO more solid solution instead of substitutions? At the end if it works for you, fine for me.

phil-davis:
@xoxys yes, now that I see it in ocis, that system is nicer than "majic" bash variable construction with double-$-sign and...

I adjusted this PR. Please review.

CC @micbar

xoxys (approved):

phil-davis:
The RocketChat notification at the end of the merge commit CI worked - good.
https://drone.owncloud.com/owncloud/activity/3454/12/1

Issue #1193, opened 2023-08-18
[QA] hard failure `apiActivity/list.feature:1254 test` in nightly


PrajwolAmatya:
Build link: https://drone.owncloud.com/owncloud/activity/3525/27/10

```feature
Scenario: adding a ownCloud server to the public link shows the share activity for the receiver     # /var/www/owncloud/testrunner/apps/activity/tests/acceptance/features/apiActivity/list.feature:1254
    Given using server "REMOTE"                                                                       # FeatureContext::usingServer()
    And user "Brian" has been created with default attributes and without skeleton files              # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And using server "LOCAL"                                                                          # FeatureContext::usingServer()
    And user "Alice" has been created with default attributes and without skeleton files              # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And user "Alice" has uploaded file with content "ownCloud test text file 0" to "/fileToShare.txt" # FeatureContext::userHasUploadedAFileWithContentTo()
    And user "Alice" has created a public link share with settings                                    # FeatureContext::userHasCreatedAPublicLinkShareWithSettings()
      | path        | /fileToShare.txt |
      | permissions | read             |
    When using server "REMOTE"                                                                        # FeatureContext::usingServer()
    And user "Brian" adds the last public link share to the remote server using the Sharing API       # ActivityContext::userHasAddedTheLastPublicLinkShareToRemoteServer()
      Fatal error: Call to undefined method FeatureContext::getLastPublicShareToken() (Behat\Testwork\Call\Exception\FatalThrowableError)
    Then the HTTP status code should be "200"                                                         # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    And user "Brian" should not have any activity entries with type "shared"                          # ActivityContext::userShouldNotHaveAnyActivityEntriesWithType()
    When using server "LOCAL"                                                                         # FeatureContext::usingServer()
    Then the activity number 1 of user "Alice" should match these properties:                         # ActivityContext::activityWithIndexShouldMatch()
      | type             | /^shared$/                                                                                                                                                    |
      | user             | /^Alice$/                                                                                                                                                     |
      | affecteduser     | /^Alice$/                                                                                                                                                     |
      | app              | /^files_sharing$/                                                                                                                                             |
      | subject          | /^shared_link_self$/                                                                                                                                          |
[...]

PR #1194, opened 2023-08-18
[tets-only] use updated link-share methods to get data

saw-jan:
Methods to get the created link share data was changed in PR https://github.com/owncloud/core/pull/40919. I have changed the tests with the updated methods

Fixes https://github.com/owncloud/activity/issues/1193


phil-davis (approved):
Looks good if CI passes.

PR #1196, opened 2023-09-26
[tests-only] fix: drop usage of ${}

phil-davis:
## Description
In preparation of php8.2 support - see https://php.watch/versions/8.2/$%7Bvar%7D-string-interpolation-deprecated

Similar to core https://github.com/owncloud/core/pull/40995

This syntax was only in the test code, so there is no need to make sure a release happens etc.


SwikritiT (approved):
LGTM

saw-jan (approved):

PR #1218, opened 2024-08-14
test(refactor): fix nullable parameters error in nightly failure

anon-pradip:
This PR fixes nightly failures caused by Implicit nullable parameters issue
Related issue: https://github.com/owncloud/QA/issues/859

### DIFF
--- a/lib/BackgroundJob/EmailNotification.php
+++ b/lib/BackgroundJob/EmailNotification.php
@@ -61,10 +61,10 @@ class EmailNotification extends TimedJob {
 	 * @param bool|null $isCLI
 	 */
 	public function __construct(
-		MailQueueHandler $mailQueueHandler = null,
+		?MailQueueHandler $mailQueueHandler = null,
 		IUserManager $userManager,
-		IConfig $config = null,
-		ILogger $logger = null,
+		?IConfig $config = null,
+		?ILogger $logger = null,
 		$isCLI = null
 	) {
 		// Run all 15 Minutes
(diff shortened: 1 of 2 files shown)


lib/BackgroundJob/EmailNotification.php
```diff
@@ -61,10 +61,10 @@ class EmailNotification extends TimedJob {
 	 * @param bool|null $isCLI
 	 */
 	public function __construct(
-		MailQueueHandler $mailQueueHandler = null,
+		?MailQueueHandler $mailQueueHandler,
```
phil-davis, line 64:
```suggestion
		?MailQueueHandler $mailQueueHandler = null,
```

We should be able to leave `= null` as the default value. Then we do not have to change code that calls the constructor.

(same for other lines below)

anon-pradip: wow

lib/BackgroundJob/EmailNotification.php
```diff
@@ -61,10 +61,10 @@ class EmailNotification extends TimedJob {
 	 * @param bool|null $isCLI
 	 */
 	public function __construct(
-		MailQueueHandler $mailQueueHandler = null,
+		?MailQueueHandler $mailQueueHandler = null,
 		IUserManager $userManager,
-		IConfig $config = null,
-		ILogger $logger = null,
+		?IConfig $config = null,
+		?ILogger $logger,
```
phil-davis, line 67:
```suggestion
		?ILogger $logger = null,
```

phil-davis (approved):

PR #1239, opened 2026-04-13
chore(ci): pass app-name to build workflow

phil-davis:
To fix:
https://github.com/owncloud/activity/actions/runs/24326294454
```

Invalid workflow file: .github/workflows/dist.yml#L16
The workflow is not valid. .github/workflows/dist.yml (Line: 16, Col: 11): Input app-name is required, but not provided while calling.
```


phil-davis:
This PR is not needed.
The requirement for app-name has been removed in https://github.com/owncloud/reusable-workflows/pull/30

PR #1237, opened 2026-04-10
chore(ci): run acceptance tests in GitHub workflows

phil-davis:
This is where I got up to earlier today, before I started trying to get it all to work with https

The drone pipelines did work without https being set up, so probably best to continue working here to debug why some acceptance tests are failing.

PR #1242, opened 2026-04-22
chore(ci): run acceptance tests in GitHub workflows

phil-davis:
Fix an issue with some calls to the parameter format method where it was sometimes called with an array of texts. The BaseFormatter code is enhanced handle process an array of text strings to sanitize.

Enhance the acceptance test ActivityContext sendActivityGetRequest method so that it handles the case when there are no activity entries. Changes to `json_decode` in PHP8 meant that the previous failed.

Provide a new WebUI test step that checks for an expected message text anywhere within the activity list. Use it in one of the acceptance test scenarios. The test was intermittent because the combination of quick addition and then deletion of files resulted in sometimes different order of rendering of activity entries that all happened in the same second.

Skip the test for adding a public link share to another federated ownCloud (issue 800). The test can be reenabled and redesigned if the issue is addressed in future.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,3 +64,36 @@ jobs:
     with:
       app-name: ${{ needs.get-vars.outputs.app-name }}
       php-version: '8.3'
+
+  acceptance-api:
+    name: API acceptance tests
+    needs:
+      - get-vars
+    uses: owncloud/reusable-workflows/.github/workflows/acceptance.yml@main
+    with:
+      app-name: ${{ needs.get-vars.outputs.app-name }}
+      do-api-tests: true
+      server-folder: 'server'
+      federated-folder: 'federated'
+
+  acceptance-webui:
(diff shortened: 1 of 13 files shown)


phil-davis:
Note: see https://github.com/owncloud/activity/pull/1237#issuecomment-4280191378

The format/sanitizeHTML fixes messages in the log like:
```
{"reqId":"0zVqhYrdijE2aSrEiTiq","level":3,"time":"2026-04-20T10:10:42+00:00","remoteAddr":"::1","user":"Brian","app":"PHP","method":"GET","url":"\/server\/apps\/activity\/api\/v2\/activity\/all?format=json&previews=true&since=0","message":"Array to string conversion at \/var\/www\/html\/server\/apps\/activity\/lib\/Formatter\/BaseFormatter.php#34"}
````

In PHP8 the level of these messages changed from notice to warning.

DeepDiver1975 (approved):

PR #1249, opened 2026-05-27
ci: make dependabot always produce convential-commit messages

phil-davis:
Should help with PRs like #1248

I think that dependabot "guesses" after a while by examining recent commits in a repo, so it might be doing "the right thing" already in some repos.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,9 @@ updates:
         update-types:
           - minor
           - patch
+    commit-message:
+      prefix: "chore"
+      include: "scope"
 
   - package-ecosystem: npm
     directory: "/"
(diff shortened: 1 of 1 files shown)


phil-davis:
If these are the settings that we want, then it would be useful to apply this to all repos. Is there an agent that will do that for us?
@DeepDiver1975

phil-davis:
https://github.com/owncloud/activity/actions/runs/26491535305/job/78010219084?pr=1249

Getting problems with Codeberg again today.

DeepDiver1975:
If a repo is already using semantic commit messages, dependabot is recognizing this and follows this.
We need to fix only repos like this...

phil-davis:
@DeepDiver1975 I forced CI to run again, and it looks good today.

DeepDiver1975 (approved):

PR #1263, opened 2026-07-23
test: bump phpstan to v2

phil-davis:
PHPstan is updated from v1 to v2. v2 is the current version supported on PHP 8.

The format of ignored errors in phpstan.neon is changed to use the new rawMessage format. Previously the messages had to be encoded as regex strings. For oridnary text messages, the rawMessage form is much simpler.

Some ignored messages were moved into the code as phpstan-ignore comment lines. This associates the ignored message directly with the exact place in the code that generates the message. That will be helpful if/when code refactoring is being done to "fix" the code that generates the phpstan message.

The PHP unit tests were emitting some PHP deprecation messages. Code has been minimally refactored to avoid these.

There are no real production run-time fixes to make here. The purpose of this PR is to demonstrate the type of minimal changes that can be needed when updating from PHPstan v1 to v2. No rush to review or merge this.

### DIFF
--- a/lib/Command/SendEmails.php
+++ b/lib/Command/SendEmails.php
@@ -81,14 +81,11 @@ class SendEmails extends Command {
 		do {
 			$users = $this->mqHandler->getAllUsers(self::BATCH_SIZE);
 			$batchCount = \count($users);
-			if ($batchCount === 0) {
-				// queue is empty
-				break;
-			}
-
-			$this->sendBatch($users, $output);
-			if ($progress !== null) {
-				$progress->advance($batchCount);
+			if ($batchCount > 0) {
+				$this->sendBatch($users, $output);
+				if ($progress !== null) {
+					$progress->advance($batchCount);
+				}
 			}
 		} while ($batchCount > 0);
 
--- a/lib/Controller/OCSEndPoint.php
+++ b/lib/Controller/OCSEndPoint.php
@@ -56,7 +56,7 @@ class OCSEndPoint {
 	protected $objectId;
 
 	/** @var string */
-	protected $user;
+	protected $user = '';
 
 	/** @var bool */
 	protected $loadPreviews;
--- a/lib/PlainTextParser.php
+++ b/lib/PlainTextParser.php
@@ -53,6 +53,8 @@ class PlainTextParser {
 	 * @return string
 	 */
 	protected function parseCollections($message) {
+		// @phpstan-ignore nullCoalesce.variable
+		$message = $message ?? '';
 		return \preg_replace_callback('/<collection>(.*?)<\/collection>/', function ($match) {
 			$parameterList = \explode('><', $match[1]);
 			$parameterListLength = \sizeof($parameterList);
--- a/lib/Data.php
+++ b/lib/Data.php
@@ -85,6 +85,7 @@ class Data {
 	 * @return bool
 	 */
 	public function send(IEvent $event) {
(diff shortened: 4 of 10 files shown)


lib/Command/SendEmails.php
```diff
@@ -81,14 +81,11 @@ public function execute(InputInterface $input, OutputInterface $output): int {
...
-			}
-
-			$this->sendBatch($users, $output);
-			if ($progress !== null) {
-				$progress->advance($batchCount);
+			if ($batchCount > 0) {
+				$this->sendBatch($users, $output);
+				if ($progress !== null) {
+					$progress->advance($batchCount);
+				}
```
phil-davis, line 88: phpstan reported that `$batchCount` was always greater than 0 when evaluated at the `while` below.
That was true, because the (only) way out of this loop was the early `break` above.

I have refactored this loop so that `while ($batchCount > 0)` is actually the place where the decision to exit the loop is made.

lib/Controller/OCSEndPoint.php
```diff
@@ -56,7 +56,7 @@ class OCSEndPoint {
 	protected $objectId;
 
 	/** @var string */
-	protected $user;
+	protected $user = '';
```
phil-davis, line 59: There were potential ways to access `$user` when it was still null.
It is supposed to be a string, so set it to the empty string at instantiation.

(Some day we could really declare all the types, and enable strict_types, etc.)

lib/PlainTextParser.php
```diff
@@ -53,6 +53,8 @@ public function parseMessage($message) {
 	 * @return string
 	 */
 	protected function parseCollections($message) {
+		// @phpstan-ignore nullCoalesce.variable
+		$message = $message ?? '';
```
phil-davis, line 57: The unit tests managed to call down to here with `$message` null.
That reported:
```
PHPUnit 9.6.35 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.32
Configuration: ./phpunit.xml
Warning:       No code coverage driver available

...............................................................  63 / 398 ( 15%)
............................................................... 126 / 398 ( 31%)
............................................................... 189 / 398 ( 47%)
............................................................... 252 / 398 ( 63%)
............................................................... 315 / 398 ( 79%)
......PHP Deprecated:  preg_replace_callback(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /home/phil/git/owncloud/core/apps-external/activity/lib/PlainTextParser.php on line 56
PHP Deprecated:  preg_replace_callback(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /home/phil/git/owncloud/core/apps-external/activity/lib/PlainTextParser.php on line 56
PHP Deprecated:  preg_replace_callback(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /home/phil/git/owncloud/core/apps-external/activity/lib/PlainTextParser.php on line 56
PHP Deprecated:  preg_replace_callback(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /home/phil/git/owncloud/core/apps-external/activity/lib/PlainTextParser.php on line 56
......................................................... 378 / 398 ( 94%)
....................                                            398 / 398 (100%)

Time: 00:03.787, Memory: 40.00 MB

OK (398 tests, 1213 assertions)
```

`$message` is supposed to be a string. But we don't explicitly declare that.
So if it is passed as null, then set it to the empty string.

